PR #49, opened 2024-05-08
Add web app support

medinnna:
Hi! I added the necessary changes for web app support:

- App icon
- Theme color matching using `background-color` from `glance.yaml`
- Navigation bar position adjustment only for iOS to avoid home indicator
- Manifest file for basic configuration
- Tested on iOS, Chrome and Safari

iOS:

ios

Chrome:

chrome

Safari:

safari

### DIFF
--- a/internal/assets/static/main.css
+++ b/internal/assets/static/main.css
@@ -1009,6 +1009,10 @@ body {
         --safe-area-inset-bottom: env(safe-area-inset-bottom, 0);
     }
 
+    .list-collapsible-label:has(.list-collapsible-input:checked) {
+        bottom: calc(var(--mobile-navigation-height) + var(--safe-area-inset-bottom));
+    }
+
     .mobile-navigation {
         transform: translateY(calc(100% - var(--mobile-navigation-height) - var(--safe-area-inset-bottom)));
         padding-bottom: var(--safe-area-inset-bottom);


internal/assets/static/main.css
```diff
@@ -1022,6 +1022,14 @@ body {
     }
 }
 
+@media (display-mode: standalone) {
+  @supports (-webkit-touch-callout: none) {
+    .mobile-navigation {
+      bottom: 15px;
```
svilenmarkov, line 1028: I don't have an iPhone so I can't test this, however could we use something like this?

```css
env(safe-area-inset-bottom, 20px);
```

Docs: https://developer.mozilla.org/en-US/docs/Web/CSS/env

If it works, it should provide a more flexible approach to offsetting the mobile navigation. Might require some extra work as well to get the "SHOW LESS" labels to also account for this offset since they get stickied to the bottom.

svilenmarkov (commented):
Hey, thanks for contributing! 

Just a small query about the mobile navigation offset since there might be a more dynamic way of doing it.

internal/assets/static/main.css
medinnna: [](url)Done! Now the safe-area variables are used, but because of this I went a little further since the way of hiding the page bar caused it to be exposed due to the offset, that's why I made some adjustments to compensate for it, I hope you don't mind. This is the final result:

iOS:


https://github.com/glanceapp/glance/assets/8866451/e1996ec9-fb2e-40e6-a773-b3d71f3badc7


Chrome:


https://github.com/glanceapp/glance/assets/8866451/640858d4-9f7f-43a8-b54b-69e65d36cabf

svilenmarkov: Thanks for looking into it! There's a high likelihood that more stuff will end up going there in the future so a static height that only accounts for the navigation links won't work. I've changed the implementation so that it can work with content of any height:

(fake navigation bar btw)

https://github.com/glanceapp/glance/assets/7613769/06fc9d60-f62c-46c9-9a37-477948223a62

Let me know if this works and I'll get it merged.

Issue #43, opened 2024-05-05
[REQUEST] Custom sorting for Reddit, HackerNews


knhash:
Add option to choose custom sorting of the feed URL. 

For instance, using the url `https://news.ycombinator.com/front` instead of `https://news.ycombinator.com` gives top posts of past day in HackerNews

Similarly for Reddit

svilenmarkov:
Hey, thanks for the suggestion. 

It's on my to-do list and I'm planning on having it released for `v0.4.0`.

stayupthetree:
I was coming to see about having my reddit results be by new instead of the default, does this fall under the above?

svilenmarkov:
@stayupthetree It does, and it is now possible to do in `v0.4.0`.

PR #56, opened 2024-05-12
Add Lobsters widget

jonasknobloch:
I threw together a widget for https://lobste.rs. ~The initial list of posts is fetched via RSS. Post metadata is retrieved by appending `.json` to the discussion URL.  Retrieval of metadata seems to be heavily rate limited.  Since we can get title and URL from the RSS feed alone, and posts are already ranked, we should be able to display the posts without metadata if necessary.~

Lobsters posts are usually tagged. For now, I simply append the tag list to the title. Maybe it makes sense to extend the forum-post template in the future for something cleaner. The feed itself can be filtered by using a feed URL like `https://lobste.rs/t/go.json` or even `https://lobste.rs/hottest.json?token={TOKEN}` instead of the default `https://lobste.rs/hottest.json`. 

Let me now what you think and I will update my patch accordingly and add the missing documentation.

Cheers!

Screenshot 2024-05-12 at 17 22 11


jonasknobloch:
Just noticed that the whole feed can be retrieved as JSON in a single request. Let me fix that 🫡

jonasknobloch:
Regarding issue https://github.com/glanceapp/glance/issues/43:

Right now I am not running `posts.CalculateEngagement()` & `posts.SortByEngagement()` to preserve the original sorting from https://lobste.rs/active. Sorting by engagement might make sense when using https://lobste.rs/recent though.

svilenmarkov:
Hey, thanks for contributing! 

I actually didn't know about lobsters until now, I'll have a look at the API in the next few days and see if I can come up with a more tailored way of adding the tags. 

Also, don't worry about the engagement sorting, it's no longer the default for Reddit or Hacker News.

svilenmarkov:
Sorry it took a while to get this merged and thanks again for contributing.

PR #75, opened 2024-05-17
Allow some branding customization.

realdavidops:
I wanted to be able to configure and disable some of the currently default branding. This pull requests adds a new top level configuration object for branding, along with several sub objects. This is tested and running fine locally for me, input is appreciated. 

## Branding
You can adjust the various parts of the branding through a top level `branding` property. Example:

```yaml
branding:
  show: true
  name: Glance
  short-name: G
```

### Properties

| Name | Type | Required | Default |
| ---- | ---- | -------- | ------- |
| show | bool | no | true |
| name | string | no | Glance |
| short-name | string | no | G |

#### `show`
True will show the glance footer, false will hide it.

#### `name`
Sets the name presented after the page name in the title.

#### `short-name`
Sets the name presented before the pages in the header.


realdavidops:
@svilenmarkov Any feedback?

svilenmarkov:
Hey, thanks for contributing and sorry I didn't get back to you earlier!

There's a few small things I'd like to change like replacing `short-name` with `logo` as I'm sure most people would rather have an image there instead of a character. I'll get it sorted out and merged for the next release.

realdavidops:
That makes sense to me. I would maybe also give a logo-type option to allow people to select either text or an image. Deciding to load text only would keep the page size down for some use cases.

vishalkadam47:
Does this include footer section also

svilenmarkov:
@vishalkadam47 Yup, I've added that as well. 

Thanks again for contributing and sorry it took so long to get this merged!

vishalkadam47:
@svilenmarkov add one option for PWA app name

PR #76, opened 2024-05-18
Add custom sorting to the twitch channels widget

fawni:
this allows the channels to be sorted as defined in the config while still keeping the live channels on the top. the wording could probably be improved, "custom" is too broad but i'm not sure what else to call it.


svilenmarkov:
Hey, thanks for contributing!

I changed `custom` to `live` since I think that more accurately conveys live channels will still take precedence over the ordering specified in the config.

PR #81, opened 2024-05-18
Specify the correct method in the Monitor widget documentation

yardenshoham:
After 189b8895b03ec2ee407c8d951c07230bae3a9a07 we use `GET` instead of `HEAD`


svilenmarkov:
Thanks for the correction! :D

PR #83, opened 2024-05-18
Image build rework

wfg:
This change makes it simpler and more convenient to build the app.

If I should use a release branch as the base, please let me know, and I will rebase.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+# https://docs.docker.com/build/building/context/#dockerignore-files
+# Ignore all files by default
+*
+
+# Only add necessary files to the Docker build context (Dockerfiles are always included implicitly)
+!/internal/
+!/go.mod
+!/go.sum
+!main.go
--- a/Dockerfile.single-platform
+++ b/Dockerfile.single-platform
@@ -1,7 +1,14 @@
+FROM golang:1.22.3-alpine3.19 AS builder
+
+WORKDIR /app
+COPY . /app
+RUN CGO_ENABLED=0 go build .
+
+
 FROM alpine:3.19
 
 WORKDIR /app
-COPY build/glance /app/glance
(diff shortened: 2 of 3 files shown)


svilenmarkov:
Hey, thanks for contributing!

wfg:
@svilenmarkov I just realized this may break your build-and-ship script.

A proper build pipeline would be a good replacement for it anyway. If you'd like, I can work on that.

svilenmarkov:
@wfg It shouldn't break the build script as it uses the other Dockerfile which wasn't changed by this PR, unless you have something else in mind? 

As for replacing it, potentially further down the line as I'm not yet ready to embrace the chaos of Makefiles, bash scripts and large Dockerfiles. Thank you for offering though!

wfg:
>It shouldn't break the build script as it uses the other Dockerfile which wasn't changed by this PR, unless you have something else in mind?

I'm only thinking about the `.dockerignore` file not including the `/build/` directory referenced here: https://github.com/glanceapp/glance/blob/main/Dockerfile#L8
The quick-and-dirty answer is to add `!/build/` to `.dockerignore`.


>As for replacing it, potentially further down the line as I'm not yet ready to embrace the chaos of Makefiles, bash scripts and large Dockerfiles. Thank you for offering though!

I was going to make GitHub Actions do all that work for us :)
I'll play around with it on my fork and may submit a draft PR for you to take a peek at. I suspect it'll be a lot simpler than you're thinking.

Awesome project, by the way! I've been trying to find something like this and happened to come across it on Hacker News.

svilenmarkov:
Ah right, yup, it does actually break it, nice catch.

If you do spend some time automating the build with GitHub Actions and it ends up being relatively straightforward while achieving much of the same then feel free to submit a PR and I'll have a look. Thanks again for offering help with this.

Also glad to hear you're finding it useful!

PR #87, opened 2024-05-19
Fix column overflow

### LINKED ISSUE #86: Text overflowing
So I just noticed that in my reddit sidebar the text was overflowing going outside the designated spot. Here is a screenshot: 

image

*It is always programmer humor lol*

hex-developer:
Fixes #86

This PR adds a `max-width-100` class, so flex items in lists do not overflow.

Here are 2 screenshots of before and after:


svilenmarkov:
Hey, thanks for looking into it!

The problem in that issue though was in the RSS widget rather than the forum posts template. 😅

PR #100, opened 2024-05-22
feat: support bilibili

teemozhu:
add a bilibili widget

it's look like Video Widget

image


svilenmarkov:
Hey, thanks for contributing!

While I would like to support more sources, I think rather than having different widgets for each source it would make more sense to have a single videos widget that supports multiple sources. I'll spend some time thinking about the best way to do this as I'd have to make some backwards incompatible changes.

helloteemo:
I agree with you that the current Bilibili Channel implementation also reuses video widgets

The difficulty with widgets that support multiple sources (including all that require network requests) is how to process responses to HTTP requests into a standard structure. Glance must implement mainstream channels, and when it encounters the rest of the channels, it leaves the processing to itself.


Maybe [gopher-lua](https://github.com/yuin/gopher-lua) would help.

jinghuashang:
Brother, this is very cool, but I want to know how to support blilibili, but I feel that my brother may be Chinese, if it is Chinese, then talk in Chinese, I am not very good in English

9k001:
> I agree with you that the current Bilibili Channel implementation also reuses video widgets
> 
> The difficulty with widgets that support multiple sources (including all that require network requests) is how to process responses to HTTP requests into a standard structure. Glance must implement mainstream channels, and when it encounters the rest of the channels, it leaves the processing to itself.
> 
> Maybe [gopher-lua](https://github.com/yuin/gopher-lua) would help.

hi 
What is the current way to support bilibili?

PR #134, opened 2024-06-07
Use GitHub's latest release API endpoint

wfg:
The current releases widget uses the releases endpoint to pull the 10 most recent releases and filter them to find the latest release. This causes a problem when a repository's latest release is outside of the 10 most recent (e.g. 10 prereleases):

```
ERROR No live release found repository=cross-seed/cross-seed url="https://api.github.com/repos/cross-seed/cross-seed/releases?per_page=10"
```

This is no longer a problem when using the latest release endpoint which grabs the latest release, ignoring draft releases and prereleases.

### DIFF
--- a/internal/feed/github.go
+++ b/internal/feed/github.go
@@ -8,12 +8,10 @@ import (
 	"time"
 )
 
-type githubReleaseResponseJson struct {
+type githubReleaseLatestResponseJson struct {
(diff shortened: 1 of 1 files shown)


svilenmarkov:
Thanks for sorting that out!

I couldn't find that endpoint in the docs when looking for it and didn't think to try `latest`. 😅

PR #137, opened 2024-06-09
feat: added property new-tab to type: search

ralphocdol:
I'm used to having a search bar on a dashboard that opens in a new tab and I kinda prefer it (muscle memory?). Although there is a keyboard shortcut for it, I tend to forget it. This is just a property that is `false` by default and switch the shortcuts if `true`.

This is also my first time with Go, so please do change and optimize things if need be.


svilenmarkov:
Thanks for contributing! 

I changed the implementation a bit to keep things consistent with how other data gets passed, other than that - good job. 🙂

PR #145, opened 2024-06-12
Rss feed enhance

ssrtw:
Fixed an issue where RSS items without provided titles, such as those from [mastodon](https://mastodon.social/@Mastodon.rss), would not show titles in the RSS list, leading to an absence of navigable links. According to the [RSS specification](https://www.rssboard.org/rss-profile#element-channel-item-title), it is permissible to omit a title. However, this led to a situation where titles were not displayed in the RSS widget, making it impossible to access the corresponding links.

Additionally, the logic for hide-categories and hide-description has been adjusted, so that these settings only affect the display when the style is set to detailed-list.

Before:
image
After:
image

Example config:
```yaml
pages:
  - name: Test
    columns:
      - size: full
        widgets:
          - type: rss
            limit: 10
            feeds:
              - url: https://mastodon.social/@Mastodon.rss
          - type: rss
            style: detailed-list
            limit: 10
            feeds:
              - url: https://mastodon.social/@Mastodon.rss
      - size: full
        widgets:
          - type: rss
            style: horizontal-cards-1
            limit: 10
            feeds:
              - url: https://mastodon.social/@Mastodon.rss
          - type: rss
            style: horizontal-cards-2
            limit: 10
            feeds:
              - url: https://mastodon.social/@Mastodon.rss
```

### DIFF
--- a/internal/feed/rss.go
+++ b/internal/feed/rss.go
@@ -93,7 +93,6 @@ func getItemsFromRSSFeedTask(request RSSFeedRequest) ([]RSSFeedItem, error) {
 
 		rssItem := RSSFeedItem{
 			ChannelURL: feed.Link,
-			Title:      item.Title,
 		}
 
 		if request.ItemLinkPrefix != "" {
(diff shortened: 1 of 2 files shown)


svilenmarkov:
Thanks for contributing!

PR #164, opened 2024-07-02
fix: allow custom bangs without an argument

### LINKED ISSUE #163: Allow custom bangs with no queries on search widget
Currently let's say I have a custom bang `!jenkins`. The search won't allow me to press enter untill I write some query in front of it like `!jenkins 123`. Which is kind of unnecessary because bangs could be without queries as well.

So I would suggest to have the flow like this, when you enter something in the search bar
- Check if it's a part of custom bangs, if yes replace it with URL
- Otherwise go to your search engines, Google,duckduckgo etc.

I am happy to contribute, just let me know if what I am saying makes sense

dracarys18:
Fixes #163

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -40,7 +40,6 @@ pages:
     columns:
       - size: small
         widgets:
-          - type: clock
           - type: calendar
 
           - type: rss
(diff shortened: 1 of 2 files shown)


svilenmarkov:
Thanks for contributing!

PR #165, opened 2024-07-03
releases: Add support for gitlab

Fumesover:
Allow the `release` module to target both `github` and `gitlab` by making it target a generic "forge" (to allow adding other git forges later)

The default source is set as github and it won't cause any issues for existing install when upgraded.


svilenmarkov:
Hey, thanks for contributing!

I think rather than having to choose a single source and adding multiple widgets for each source, what might make more sense is to be able to choose the source for each repository individually. So, rather than this:

```yaml
- type: releases
  repositories:
    - immich-app/immich

- type: releases
  source: gitlab
  repositories:
    - inkscape/inkscape
```

Something like this which defaults to GitHub but allows you to specify a prefix for other sources:

```yaml
- type: releases
  repositories:
    - immich-app/immich
    - gitlab:inkscape/inkscape
```

PR #168, opened 2024-07-11
replaced markets to stocks

esh-1:


svilenmarkov:
The widget was intentionally renamed to `markets` in `v0.5.0`. If you made this change because you were having issues please try using the latest version.

PR #171, opened 2024-07-24
Fix incorrect LiveSince if API returned no game

DVDAndroid:
Fix incorrect LiveSince if API returned no game.

Streamer can stream without a category/game associated.

streamer


svilenmarkov:
Thanks for contributing!

PR #172, opened 2024-07-24
Make Twitch avatar clickable

DVDAndroid:


svilenmarkov:
Thanks for contributing!

PR #173, opened 2024-07-25
Fix crash when using a custom CSS file; fix error 404 for manifest.json

DVDAndroid:
In release 0.6.0, when using a custom CSS file, backend crashes with error `template: page.html:17:76: executing "document-head-after" at <.App.Config.Server.StartedAt.Unix>: can't evaluate field StartedAt in type glance.Server`.

Also fixed error 404 for manifest.json

### DIFF
--- a/internal/assets/templates/document.html
+++ b/internal/assets/templates/document.html
@@ -13,7 +13,7 @@
     <meta name="theme-color" content="{{ if ne nil .App.Config.Theme.BackgroundColor }}{{ .App.Config.Theme.BackgroundColor }}{{ else }}hsl(240, 8%, 9%){{ end }}">
     <link rel="apple-touch-icon" sizes="512x512" href="{{ .App.AssetPath "app-icon.png" }}">
     <link rel="icon" type="image/png" sizes="50x50" href="{{ .App.AssetPath "favicon.png" }}">
-    <link rel="manifest" href="/static/manifest.json">
+    <link rel="manifest" href="{{ .App.AssetPath "manifest.json" }}">
     <link rel="icon" type="image/png" href="{{ .App.AssetPath "favicon.png" }}" />
     <link rel="stylesheet" href="{{ .App.AssetPath "main.css" }}">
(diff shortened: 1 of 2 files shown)


DVDAndroid:
Why `Config.Server.StartedAt` has been removed? Did you want to change some behavior?

Out of curiosity, why are you managing assets with a subfolder `<hash>` ?

svilenmarkov:
> Out of curiosity, why are you managing assets with a subfolder \<hash\> ?

I've been wanting to add more features that rely on JS but I think the single JS file approach is becoming unmanageable so I'm in the process of splitting things into multiple files. I want to use JS's native `import` but doing so means I can't use query parameters when importing files, so the `StartedAt` approach is a no go. I can however use relative paths to achieve the same result. If in `main.js` I have:

```js
import someFunction from "./utils.js"
```

and `main.js` is in `/static/<hash>/main.js`, then `utils.js` will be imported from `/static/<hash>/utils.js`. Because the hash is computed based on the contents of all asset files, that also means that restarting the server won't force you to re-download the same assets if they haven't changed, like it does in v0.5.0. 

The manifest and custom CSS are somewhat special edge cases that don't fit this model very well but I think we can go with these fixes for now until I find a better solution.

DVDAndroid:
> The manifest and custom CSS are somewhat special edge cases that don't fit this model very well but I think we can go with these fixes for now until I find a better solution.

Or you can concatenate path with `{{ .App.AssetsHash }}` instead of using `{{ .App.AssetsPath }}`, removing `StartedAt`.

Like:
`<link rel="stylesheet" href="{{ .App.Config.Theme.CustomCSSFile }}?v={{ .App.AssetsHash }}">`

If I am understanding this correctly

svilenmarkov:
That wouldn't work since the custom CSS file is served from the user-specified assets directory, which is separate from the static server assets that the hash gets computed from. This means that if you were to edit the file and restart the server you'd still get served the same file until the cache expires (or you manually clear it). The static server files are embedded in the binary where as the user assets get served using `http.FileServer`.

Issue #106, opened 2024-05-24
[FEATURE] Youtube channels exclude shorts from showing


bigsk1:
Not sure if there is a way to not see shorts from youtube channels and only actual videos.

helloteemo:
Maybe the YouTube RSS link won't be added to shorts

bigsk1:
an idea

extract the ytp-time-duration from html

Add function

```
package feed

import (
	"fmt"
	"io"
	"net/http"
	"regexp"
	"strings"
	"time"
)

// fetchVideoDurationFromHTML fetches the video duration by parsing the HTML of the video page.
func fetchVideoDurationFromHTML(videoURL string) (time.Duration, error) {
	resp, err := http.Get(videoURL)
	if err != nil {
		return 0, err
	}
	defer resp.Body.Close()

	body, err := io.ReadAll(resp.Body)
	if err != nil {
		return 0, err
	}

	// Regular expression to find the duration in the HTML
	re := regexp.MustCompile(`class="ytp-time-duration">(\d+):(\d+)</span>`)
	matches := re.FindStringSubmatch(string(body))
	if len(matches) < 3 {
		return 0, fmt.Errorf("duration not found")
	}

	// Parse minutes and seconds
	minutes, err := strconv.Atoi(matches[1])
	if err != nil {
		return 0, err
	}
	seconds, err := strconv.Atoi(matches[2])
	if err != nil {
		return 0, err
	}

	duration := time.Duration(minutes)*time.Minute + time.Duration(seconds)*time.Second
	return duration, nil
}
```
Update FetchYoutubeChannelUploads

```
package feed

import (
	"fmt"
	"log/slog"
	"net/http"
	"net/url"
	"strings"
	"time"
)

type youtubeFeedResponseXml struct {
	Channel     string `xml:"title"`
	ChannelLink struct {
		Href string `xml:"href,attr"`
	} `xml:"link"`
	Videos []struct {
		Title     string `xml:"title"`
		Published string `xml:"published"`
		Link      struct {
			Href string `xml:"href,attr"`
		} `xml:"link"`

		Group struct {
			Thumbnail struct {
				Url string `xml:"url,attr"`
			} `xml:"http://search.yahoo.com/mrss/ thumbnail"`
		} `xml:"http://search.yahoo.com/mrss/ group"`
	} `xml:"entry"`
}

func parseYoutubeFeedTime(t string) time.Time {
[...]

PR #176, opened 2024-08-02
feat: add no_shorts option for YouTube feeds

3rd:
This adds a `no_shorts` option for YouTube feeds.
It works by using the default video playlist feed instead of the channel feed.
I find it very hard to express how much I hate YouTube shorts.

It would be cleaner to add support to "playlists" along with "channels", so that users could add whatever built-in/custom playlist they wish, but  I feel wanting to remove shorts is something so common that it should be handled automatically.

Related issues (pinging some other people):
- https://github.com/glanceapp/glance/issues/106
- https://github.com/FreshRSS/Extensions/issues/234
- https://github.com/Ranchero-Software/NetNewsWire/issues/3975

### DIFF
--- a/internal/feed/youtube.go
+++ b/internal/feed/youtube.go
@@ -78,12 +78,6 @@ func FetchYoutubeChannelUploads(channelIds []string, videoUrlTemplate string, no
 
 		for j := range response.Videos {
 			video := &response.Videos[j]
-
-			// TODO: figure out a better way of skipping shorts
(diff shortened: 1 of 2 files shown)


svilenmarkov:
No idea how you figured this out but thanks, this is the silver bullet I've been looking for! 

I actually went ahead and inverted the property since I think the shorts being excluded is the desired outcome in the vast majority of cases.

Issue #154, opened 2024-06-26
Feature Request - Different URLs for monitor widget


MrExplode:
I would like to set different URLs for status checking and the UI link, something like this:
```yml
- type: monitor
  cache: 1m
  title: Services
  sites:
    - title: Jellyfin
      check: https://jellyfin.yourdomain.com/health
      link: https://jellyfin.yourdomain.com
      icon: /assets/jellyfin-logo.png
```

PR #177, opened 2024-08-03
Different monitor urls

MrExplode:
Add different monitor URL for UI link and actual status checking. See corressponding issue: #154 

I would like to receive feedback on the correctness of how I handled the fallback when `check_url` is not provided, I'm not familiar how Go Yaml handles missing properties.


svilenmarkov:
Hey, thanks for contributing! 

You were correct to assume that the value, when undefined, is an empty string. The YAML package uses the default zero values for each type if not provided.

PR #181, opened 2024-08-04
Bookmark status indicator

DVDAndroid:
Implemented a status indicator for bookmarks that have a status page. For example, [https://www.githubstatus.com/](https://www.githubstatus.com/) or [https://www.redditstatus.com/](https://www.redditstatus.com/) uses the same status page dashboard and API ([statuspage.io](statuspage.io))

Updated documentation; here's a preview:
bookmarks-statuspages

Status page is an optional property. If the API returns that the page is operational, it's possible to hide the indicator (empty dot); else the indicator will be fully red (if major/critical issue) or empty with red border (if minor issue). *used per-existing indicators*


DVDAndroid:
Marking as draft because it's working in dev, but not in release mode...

PR #212, opened 2024-09-09
added support for both channel and playlist IDs

vishalkadam47:
- Modified function signature to accept both channel and playlist IDs
- Added logic to differentiate between channel IDs and playlist IDs
- Implemented direct support for playlist IDs in feed URL construction
- Enhanced error logging for better debugging and user feedback

### DIFF
--- a/internal/feed/youtube.go
+++ b/internal/feed/youtube.go
@@ -37,16 +37,18 @@ func parseYoutubeFeedTime(t string) time.Time {
 	return parsedTime
 }
 
-func FetchYoutubeChannelUploads(channelIds []string, videoUrlTemplate string, includeShorts bool) (Videos, error) {
-	requests := make([]*http.Request, 0, len(channelIds))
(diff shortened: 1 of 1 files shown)


svilenmarkov (commented):
Thanks for contributing!

It appears you have disabled the permission that allows maintainers to edit PRs so I can't make the change I wanted to, which is specifying playlists via a `playlist:` prefix rather than automatically promoting IDs starting with `PL` to a playlist. The reason for this change is that as stated [here](https://github.com/glanceapp/glance/issues/214#issuecomment-2405905665), not all playlists start with that prefix so it would be nice to allow users to specify non-standard playlist IDs. 

I've gone ahead and [pushed that change to the v0.7.0 branch](https://github.com/glanceapp/glance/commit/e3bdc730134ae47f51715cb26ca136aa06ab2012), so I'll close this PR.

PR #220, opened 2024-09-14
Add important details about the server.base-url property

paricbat:
I was a bit confused while setting up my own instance about having Glance under a subdirectory in a reverse proxy, so this PR clears it up :)


svilenmarkov (approved):
Good point, hadn't thought about that. Thanks for contributing!

PR #224, opened 2024-09-19
DNS Stats widget - fix Pihole privacy edge case

SimJunYou:
There are privacy settings on the Pihole which affect the output of the API.
pihole

When **any privacy settings other than "Show everything"** is selected on the Pihole, the DNS Stats widget fails with this error:

widget_with_privacy

The problem is that one of the fields (`top_ads`) in the JSON response from Pihole's API turns to an empty array because of the setting. For reference, with the privacy setting...
with_privacy

... and without:
without_privacy

I made a fix in `pihole.go` -- it just uses an alternate struct definition that does not expect `top_ads` if the original one fails. It copies to the original struct instance to sidestep any type errors, though it does look a little ugly. Here's how it looks after the fix:
after_fix

It's my first time using Go, so I'm not sure if this was the best way to fix it. Do let me know if there's anything I can do to improve this.

### DIFF
--- a/internal/feed/pihole.go
+++ b/internal/feed/pihole.go
@@ -18,6 +18,17 @@ type piholeStatsResponse struct {
 	DomainsBlocked    int            `json:"domains_being_blocked"`
 }
 
+// If user has some level of privacy enabled on Pihole, `json:"top_ads"` is an empty array
+// Use alternate struct without that field to avoid error when unmarshalling
+type piholeStatsResponsePrivate struct {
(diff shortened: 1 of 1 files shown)


svilenmarkov (approved):
Thanks for looking into this and contributing with so much info and a solution!

I've made a small tweak where instead of duplicating the entire struct I added a new type for the blocked domains with a custom `UnmarshalJSON` method that defaults to an empty map if there were any issues with unmarshalling.

Issue #226, opened 2024-09-27
Feature Request - Add a healthcheck route


xunleii:
First of all, thanks for this awesome starte page!

Is there a route for a health check that returns 200 when Glance is running? I didn't see one in the code. A possible usecase for this route would be a Kubernetes probe or when Glance is behind a loadbalancer. 

Sure, we could use the `/`, but depending on what's on the first page, that could be cumbersome to execute and consume resources for nothing.

If it didn't exists, a simple `mux.HandleFunc("GET /api/healthz", func (w http.ResponseWriter, _ *http.Request) {w.WriteHeader(http.Ok)})` should be enough.

svilenmarkov:
Thanks for the idea! I've added the endpoint in the new `v0.6.2` release.

PR #242, opened 2024-10-18
feat: HEALTHCHECK in Dockerfile

CremaLuca:
In commit 9a36187 the /api/healthz was added to resolve issue #226; I used this feature to use the Docker health check functionality.

There should be two ways to do it:
1. Add `--health-cmd` to `docker run` or `healthcheck` to `docker-compose` file
2. Add it to the Dockerfile so it is enabled by default

I went for the second option as you can change the details of the healtz endpoint without having to change every user's configuration.

I set the `timeout`, `start-period` and `interval` arbitrarily, can be changed as you like.

### Example

The healthcheck works even if external port is changed, I tested mine with port 8045.

As the container is starting the status is "starting"
```
IMAGE                              COMMAND                  CREATED         STATUS                             PORTS                                                     
cremaluca/glance:0.0.1             "/app/glance"            4 seconds ago   Up 3 seconds (health: starting)    0.0.0.0:8045->8080/tcp, [::]:8045->8080/tcp
```

But after a few seconds it has already changed to healthy
```
IMAGE                              COMMAND                  CREATED         STATUS                            PORTS                                                        
cremaluca/glance:0.0.1             "/app/glance"            9 seconds ago   Up 8 seconds (healthy)            0.0.0.0:8045->8080/tcp, [::]:8045->8080/tcp
```


svilenmarkov (approved):
Thanks for contributing!

PR #246, opened 2024-10-25
feat: add custom branding support for PWA

vishalkadam47:
Added dynamic manifest generation to support customizable PWA branding through glance.yml configuration. Users can now set custom PWA names using `pwa-text` and leverage the existing `favicon-url` property for both favicon and PWA icon. 

```
branding:
  logo-text: G
  pwa-text: Glanceapp
```


AmagiDDmxh:
lgtm

PR #248, opened 2024-10-28
fix: non-whole-hour timezones now correctly shown on Clock widget

oliver-mitchell:
### Problem

Certain timezones use non-whole-hour definitions, for example, Australia/ACDT - which is GMT +10:30. Due to this, if either:
1. The user is in a whole-hour timezone (eg: CEST GMT+1), and they add a non-whole-hour timezone, or
2. The user is in a non-whole-hour timezone (eg: ACDT GMT+10:30), and they add a whole-hour timezone, 

then the relative difference between timezones is displayed within the Clock widget incorrectly, where the minutes are ignored, and only hourly differences being shown. Using the example of the user being in CEST, and requesting that ACDT be shown, only `+9h` will be displayed.

This is especially critical in timezones that relatively differ from each other by only 30 minutes, as the clock widget will show _no difference_ in timezones, represented as `0h` (for example, ACDT GMT+10:30 and AEDT GMT+11:00).

### Solution
This fix changes how the difference is calculated to account for both minutes, and hours.

Due to the change, the text shown for relative timezone differences also had to be changed, as the previous formatting was not appropriate for showing both minutes and hours. A similar format to the original was tested, which resulted in strings such as `+12h30m` or `+12h 30m`, both of which I deemed looked a bit strange. As such, I made a formatting change to instead use a standard `+-HH:MM` across the board (ie: `+12:30`).

### DIFF
--- a/internal/assets/static/js/main.js
+++ b/internal/assets/static/js/main.js
@@ -502,9 +502,24 @@ function timeInZone(now, zone) {
         timeInZone = now
     }
 
-    const diffInHours = Math.round((timeInZone.getTime() - now.getTime()) / 1000 / 60 / 60);
+    const diffInMinutes = Math.round((timeInZone.getTime() - now.getTime()) / 1000 / 60);
 
-    return { time: timeInZone, diffInHours: diffInHours };
+    return { time: timeInZone, diffInMinutes: diffInMinutes };
+}
+
+function zoneDiffText(diffInMinutes) {
+    if (diffInMinutes == 0) {
+        return "";
+    }
+
+    const sign = diffInMinutes < 0 ? "-" : "+";
+
+    diffInMinutes = Math.abs(diffInMinutes);
+    
+    const hours = `${Math.floor(diffInMinutes / 60)}`.padStart(2, '0');
+    const minutes = `${diffInMinutes % 60}`.padStart(2, '0');
+
+    return `${sign}${hours}:${minutes}`;
 }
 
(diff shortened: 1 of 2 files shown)


svilenmarkov (approved):
Thanks for contributing! I see what you mean about `+12h30m` and `+12h 30m` looking odd, both looked a bit "too much". I think the diff text isn't all that important hence the deemphasized color and why I had it rounded to the hour, though I agree it is misleading for some timezones. I reverted it to how it looked initially but now if the difference is less than an hour it shows the minutes and if it's an hour and some minutes it has a `~` suffix. You can now also hover over the text to see the exact difference which should clear things up.

oliver-mitchell:
> Thanks for contributing! I see what you mean about `+12h30m` and `+12h 30m` looking odd, both looked a bit "too much". I think the diff text isn't all that important hence the deemphasized color and why I had it rounded to the hour, though I agree it is misleading for some timezones. I reverted it to how it looked initially but now if the difference is less than an hour it shows the minutes and if it's an hour and some minutes it has a `~` suffix. You can now also hover over the text to see the exact difference which should clear things up. Here's what that looks like:
> 
> image

@svilenmarkov my pedantic-nesss would looove to have the full time listed, but I appreciate the simpler format. Is this something that I could put a PR in to add a config item that would determine the formatting? I think a lot of people would also appreciate the option to have [+-]HH:MM formatting

PR #250, opened 2024-10-30
Add title prop to Market Name for full text of truncated names on hover

ehaughee:
Added a HTML `title` param to the Markets Name `div` so that when longer names are used, and truncated, the full value can be seen on hover.
market-name-hovertext


ehaughee:
Accidentally included other commits.  Will open another PR.

Issue #229, opened 2024-10-02
New Feature: Search bar acts like address bar


colinslegacy:
It'd be nice if the search bar could act like the address bar like modern browsers, with the ability to search or go to a URL by default.

Right now I have it setup where the default behavior going to a URL, then I use a bang "!s" to search, but it'd be nice to be able to eliminate the bang and set a default search engine for search alongside being able to input URLs.

svilenmarkov:
Thanks for the suggestion, I'll get this implemented for the next release.

develoopeer:
I think this feature request covering only opening sites if they are URL, to achieve modern browsers behavior, but not about  providing any search suggestions.

For more info you can check #256 Pull Request.

colinslegacy:
For this feature request, yes I'm only referring to opening URLs.  @ryaneford search suggestions would be a great idea as well, though.

PR #256, opened 2024-11-14
Search bar acts like address bar feature implemented

develoopeer:
# Feature from #229 issue implemented with the following behavior:
If string passed from input is a valid URL, tab with this string will be opened, not the search engine with URL as a search request.

URL from search input field will be considered **valid**, if they are corresponding to [RFC 3886](https://www.rfc-editor.org/rfc/rfc3986), Generic Syntax of URI`s. That does mean that string:
- http://google.com/ is valid
- https://google.com/ is valid
- google.com - is not valid


svilenmarkov:
Hey, thanks for contributing!

I'm not sure there's any benefit to strictly following the URI RFC, as in my browser's address bar I can simply type `google.com` which would be treated as a valid address. Plus, I don't think anyone would actually want to fully type `https://` every time. 

Something like the following regex should probably suffice in returning `true` for `isValidUrl` as a backup when the URL instantiation fails:

```
^(?:[a-z0-9-]+\.)+[a-z]+$
```

develoopeer:
The benefit of following the URI RFC is the fact, that you can pass URI to javascript `window.open` function directly.
If string like `google.com` will be passed to it, address of current page will be prepend to URL, and you will get `localhost/google.com`. You need to prepend a scheme to string by yourself.

But i feel like you noticed correctly. No one would type `https://` manually, or even if you have URL string in OS buffer, you will pass it to your browser, not to the search widget.

Issue #208, opened 2024-09-08
[Request] Please trigger opening of the keyboard on Android when autofocus is on for a search tab widget


driftywinds:
With v0.6.0 the autofocus functionality was released in the search widget

![image](https://github.com/user-attachments/assets/216e82cc-4aa2-4168-9a32-3fa12108c60c)

The use of this is rendered obsolete because on Android you have to tap the search bar to start typing and consequently search, regardless of whether autofocus is on or off. Please allow for the functionality of triggerring the opening of keyboard when autofocus is enabled for a search widget so that the functionality is usable for Android.

svilenmarkov:
Thanks for the suggestion! 

I'll try and see if I can get this working.

svilenmarkov:
This appears to not be possible on phones as the browsers do not allow you to programmatically invoke the keyboard.

PR #258, opened 2024-11-15
Better autofocus, to make it work on mobile somehow

develoopeer:
## Whats the point
As mentioned in issue #208, `autofocus` attribute on search bar do not works on mobile devices. This is due to security restrictions of Android/IOS devices. Keyboard cant be toggled without any kind of user interaction. 
Here's what people from react-select who have encountered the same problem [say about it](https://github.com/JedWatson/react-select/issues/3501)

I don't think there is any workaround in this case, to make this functionality works.
But this PR, adding border animation on search bar to make them works like in desktop version. Without this change, user can`t even figure out, that autofocus flag is enabled, because borders are blurred.

### DIFF
--- a/internal/assets/static/js/main.js
+++ b/internal/assets/static/js/main.js
@@ -614,6 +614,10 @@ async function setupPage() {
         setupLazyImages();
     } finally {
         pageElement.classList.add("content-ready");
+        const inputElement = document.getElementsByClassName("search-input")[0];
+		if (inputElement.dataset.autofocus == 'autofocus'){
+			inputElement.focus()
+		}
 
         for (let i = 0; i < contentReadyCallbacks.length; i++) {
             contentReadyCallbacks[i]();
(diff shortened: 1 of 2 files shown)


svilenmarkov:
Thanks for contributing!

I tried both the current implementation as well as this one and the behavior appears to be the same - the field gets focused and receives the colored border, but as you mentioned, the keyboard does not appear due to security restrictions. Is the field not getting focused for you at all with just the `autofocus` attribute?

develoopeer:
Yes the field wasn't getting any focus at all, but today i tried to reproduce this, and eventually nothing happening.
I am using Mobile version of Chrome v130.0.6. I do not know why this behavior no longer appears, maybe i was doing something wrong.
If this behavior was just false positive,  there is no need to merge this PR, because it will just make things more complicated. Sorry for bothering

PR #261, opened 2024-11-19
Docker widget

baranovskis:
I added a new widget to monitor Docker containers.
Documentation is also updated.


svilenmarkov:
Thanks for contributing!

I'll get this added in v0.8.0 since there's already a good amount of things in place for v0.7.0. 

Do you think it's feasible to change the implementation to instead make an [HTTP request to `docker.sock`](https://docs.docker.com/reference/api/engine/version/v1.47/#tag/Container/operation/ContainerList) when fetching container data, to help with keeping dependencies to a minimum?

svilenmarkov (approved):
Slight change of plans, it's going in v0.7.0. 😅 I've simplified the design a bit and added a bunch more features. Thanks again for contributing!

PR #286, opened 2024-12-24
Mobile style fixes

DallasHoff:
This pull request aims to fix a few issues with styling on mobile devices.

1. `safe-area-inset-bottom` was used to add extra padding to the navigation bar to get it out of the way of the OS gesture bar, but no corresponding padding was present on the content. This meant that some of the content was blocked at the bottom by the nav bar when scrolled all the way down.
2. Tapping the search bar widget caused the page to zoom in on iOS. I changed the font-size to 16px, the minimum font size at which iOS will not zoom the page when focusing an input.
3. When there is not enough content on a dashboard to require scrolling, the mobile nav bar did not sit at the bottom of the page (see image below), so I changed the content height from `100%` to `100dvh` to fix this.

Screenshot 2024-12-24 at 2 17 51 PM

### DIFF
--- a/internal/glance/static/main.css
+++ b/internal/glance/static/main.css
@@ -1701,6 +1701,11 @@ details[open] .summary::after {
     .weather-column-rain::before {
         background-size: 7px 7px;
     }
+
+    .search-input {
+        /* so that iOS Safari does not zoom the page when the input is focused */
+        font-size: 16px;
+    }
 }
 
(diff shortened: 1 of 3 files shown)


svilenmarkov (approved):
Thanks for contributing! I've made a few small tweaks to the implementation and also fixed an issue with the expand button not being properly positioned above the navigation.

PR #307, opened 2025-01-09
feat: Add option to include GitHub repository pre-releases in releases widget

Mystically11:
Hello

I've added the ability to show GitHub prereleases in the releases widget, as I frequently use a lot of beta versions of projects and I would find this useful.

Turned on:
image

Turned off:
image


svilenmarkov (approved):
Thanks for contributing! I slightly changed the functionality so that you can specify `include-prereleases: true` on a per-repo basis rather than for all of them.

PR #314, opened 2025-01-15
feat: add alternate link (error-url) for monitor widget sites

JeckDev:
This provides an optional parameter, error-url, per each site defined in a monitor widget. The parameter allows an alternate link to be returned when the site has an error (because the normal link by definition probably doesn't work).

For my usage, when a site returns an error it's most often because the docker is down or the site is misconfigured somehow; in either case I'd need to edit the configuration in my docker manager. So I have the error-url setup to point at a per site "http://docker-manager.local/specific-container" that's only available on my local network.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1220,6 +1220,7 @@ Properties for each site:
 | title | string | yes | |
 | url | string | yes | |
 | check-url | string | no | |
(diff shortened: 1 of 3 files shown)


svilenmarkov (approved):
Thanks for contributing!

PR #322, opened 2025-01-23
feat: add public-url parameter to rss widget feeds

JeckDev:
The rss feed widget shows a feed source which links directly to the feed url. You can find the link I'm referring to underneath each rss item's title, to the right of the "X hours ago" notice. Clicking that link in my browser downloads the rss.xml file, which is not preferred behavior. 

This provides an optional parameter `public-url` which can be set per feed to override this behavior.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -532,11 +532,15 @@ An array of RSS/atom feeds. The title can optionally be changed.
 | Name | Type | Required | Default | Notes |
 | ---- | ---- | -------- | ------- | ----- |
 | url | string | yes | | |
+| public-url | string | no | | |
 | title | string | no | the title provided by the feed | |
 | hide-categories | boolean | no | false | Only applicable for `detailed-list` style |
 | hide-description | boolean | no | false | Only applicable for `detailed-list` style |
 | item-link-prefix | string | no | | |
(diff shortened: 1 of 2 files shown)


JeckDev:
In the future, it may be worth changing current default behavior (link to the feed url) with a link to the feed's base url ("www.example.com" instead of "www.example.com/feeds/example-feed.xml"). 

I've left the default behavior unchanged here. One point against changing things, I do think some RSS feeds may include a style element which makes viewing the feed xml in browser more palatable, but I don't know any offhand to test.

svilenmarkov (approved):
Thanks for contributing!

> In the future, it may be worth changing current default behavior (link to the feed url) with a link to the feed's base url ("www.example.com" instead of "www.example.com/feeds/example-feed.xml").

The URL that is currently used for the channel is taken from the `<link>` provided by the RSS feed, which [as per the spec](https://www.rssboard.org/rss-specification#requiredChannelElements) should be:

> The URL to the HTML website corresponding to the channel.

If a feed specifies the `<link>` as the link to the feed rather than the site itself, I believe that's an issue with the feed. We could potentially check to see if `<link> == feedURL` and only then convert it to the base URL assuming no `public-url` is provided.

PR #358, opened 2025-02-17
Auto-reload config file on RENAME

rubiojr:
Some editors (like Vim), create a temp file when saving, then replace the file being edited with the temp file. This causes the FS notify event to be RENAME, not WRITE, so auto-reload misses this.

In addition to that, the file is removed from the watcher and the auto-reload functionality stops working entirely after the first RENAME.

https://github.com/fsnotify/fsnotify/issues/255 describes this.

### DIFF
--- a/internal/glance/config.go
+++ b/internal/glance/config.go
@@ -286,6 +286,20 @@ func configFilesWatcher(
 				}
 				if event.Has(fsnotify.Write) {
 					debouncedCallback()
+				} else if event.Has(fsnotify.Rename) {
+					// wait for file to be available
+					for i := 0; i < 20; i++ {
+						_, err := os.Stat(mainFileAbsPath)
+						if err == nil {
+							break
+						}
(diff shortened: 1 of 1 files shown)


svilenmarkov:
Thanks for contributing!

I've made a few small tweaks to the implementation, namely use `event.Name` instead of `mainFileAbsPath` since we also watch any included files for changes, as well as delete the path from the manually tracked includes, which will cause `debouncedParseAndCompareBeforeCallback` to add it back later if it's still needed after parsing the config. 

I _think_ that should make it a little more resilient in some horrible edge cases. Let me know if it still works for you or if you have any feedback and I'll be happy to get it merged.

rubiojr:
Thanks for the edits! I don't have a Windows around to verify, but it's working like a charm on :penguin: :bow:

svilenmarkov (approved):
Awesome, thanks for letting me know! I'll get this released as a patch for v0.7 in a bit.

Issue #316, opened 2025-01-18
Allow HTTP Basic access authentication in monitor


logoff:
Current configuration of Monitor sites do not allow HTTP Basic access authentication. It will be very useful allowing it, so some secured sites can be monitored too.

ejsadiarin:
Hey, I added a pull request in `dev` branch for this feature.  So it can be used like this:
```yaml
- title: Traefik
  url: https://traefik.example.com
  icon: di:traefik
  basic-auth:
    username: your-username
    password: your-password
```

As of now on v0.7.2, a workaround is to use `alt-status-codes` (array) and add `401` (Unauthorized) and/or `403` (Forbidden) like so:

```yaml
- title: Traefik
  url: https://traefik.example.com
  icon: di:traefik
  alt-status-codes:
    - 401
    - 403
```

PR #367, opened 2025-02-19
feat(monitor): add basic-auth feature for protected sites

ejsadiarin:
this closes [issue #316](https://github.com/glanceapp/glance/issues/316)

This feature properly resolves sites that require basic auth credentials, so that it properly returns status `OK` instead of `Unauthorized`. 

At the very least, this can replace using `alt-status-codes` and adding `401` to it, which is like a "bogus" status `OK` (although if the intent is to just ping the site if its alive, then its good enough of a solution).
```yaml
- type: monitor
  cache: 1m
  title: Services
  sites:
    - title: Traefik
      url: https://traefik.example.com
      icon: di:traefik
      basic-auth:
        username: your-username
        password: your-password
```

Furthermore, this could be expanded to also pass the configured basic auth credentials to the request when the user clicks on the specific monitor widget.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1476,6 +1476,7 @@ Properties for each site:
 | allow-insecure | boolean | no | false |
 | same-tab | boolean | no | false |
 | alt-status-codes | array | no | |
+| basic-auth | object | no | |
 
 `title`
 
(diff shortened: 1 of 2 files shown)


svilenmarkov (approved):
Thanks for contributing!

PR #382, opened 2025-03-01
Added gruvbox theme

legoraft:
Hi! I saw that the gruvbox theme was missing from the themes directory. I went ahead and added it to the themes docs for easier accessibility.


svilenmarkov (approved):
Looks nice, thanks for contributing!

PR #385, opened 2025-03-02
feat: custom-api multiple API queries

ralphocdol:
This is mainly to query multiple endpoints at once in the same `custom-api`, some services with API have the data you'd need, spread across multiple endpoints. Having only one kinda limits things so here we are.

You can also query different endpoints.

Sample config
```yaml
- type: custom-api
  define: <redacted>
  api-queries:
    yesterday:             # <--- This will become the object name where the api response gets stored
      url: http:${WEBSERVER_URL}/api/get-sonarr-data
      parameters:
        hour-offset: 8
        day-offset: -1
      <<: *sonarr-headers
    today:
      url: http:${WEBSERVER_URL}/api/get-sonarr-data
      parameters:
        hour-offset: 8
      <<: *sonarr-headers
    freshrss:
      url: http:${FRESHRSS_API}
      <<: 
        - *freshrss-parameters
        - *freshrss-headers
  cache: 1s
  template: |
    <div> - {{ .JSON.String "yesterday.data.0.airDate" }}</div>
    <div> - {{ .JSON.String "today.data.0.airDate" }}</div>
    <div> - {{ .JSON.String "freshrss.items.0.title" }}</div>
```

Output:
Screenshot_20250303_010949

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1290,34 +1290,12 @@ Examples:
 #### Properties
 | Name | Type | Required | Default |
 | ---- | ---- | -------- | ------- |
-| url | string | yes, unless `api-queries` is set | |
+| url | string | yes | |
 | headers | key (string) & value (string) | no | |
 | frameless | boolean | no | false |
 | template | string | yes | |
-| parameters | key & value | no | |
-| api-queries | list of urls, parameters & headers | no | |
-
-> [!NOTE]
-> 
-> `api-queries` will override `url`, `headers` and `parameters`
-> since it also provides its own options
-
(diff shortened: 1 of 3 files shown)


svilenmarkov (approved):
Thanks for contributing!

This is a pretty cool idea but I ended up changing the implementation and functionality a bit. It now executes all requests concurrently which should make it quite a bit faster in some cases and also allows you to access the `Response` of each subrequest as you would for the main one. 

Let me know if you have any thoughts about this change and I'd be happy to address them, otherwise I'll get this merged as is.

PR #399, opened 2025-03-07
Update configuration.md

KadoBOT:
Correct `title` to `name`.

I'm unsure if `title` is also accepted, I haven't tested it. Please close this PR if this fix is not valid.


svilenmarkov (approved):
Not sure how that managed to go unnoticed for so long, thanks. 😅 

This `name` vs `title` thing is something I've been bothered by for a while now and I might just end up making everything `name` and deprecating `title` for the sake of consistency.

PR #405, opened 2025-03-10
fix: full width clickable link for monitor-site

ralphocdol:
I always find myself clicking a blank canvas next to the actual link, this just aims to expand the clickable area

The screenshot is just to show where the clickable part is.
Before:
Screenshot_20250310_164158

After:
Screenshot_20250310_164225

This is more noticeable on mobile mode. Have to click across with just a right hand.


svilenmarkov (approved):
Thanks for contributing!

I changed the implementation to use `flex-grow: 1` instead of `width: 100%` since `min-width: 0` is required for the text truncation to work.

PR #430, opened 2025-03-15
feat: Count Timer (Reminder) widget

### LINKED ISSUE #427: Countdown / Countup timer for memorable days
### Description

I would like to see a **countdown / count up timer to/from a specific date or time**.

Personally, I'll use it to track paper submission deadlines. It can also be used for memorable days.

**Configuration** would include:
- Target date/time
- Which fields to show / Timer granularity (Days, Hours, Minutes, Seconds), or how many fields to show. For example,
  - When there's more than 100 days, show days only.
  - when there's <100days but more than 1 days, show like "10 DAYS, 23 HRS". 
  - When less than a day, show "23HRS, 10 MIN". When less than an hour, show "10 MIN, 50 SECS", etc.
  - This approach can keep the size of the widget fixed. But there might be better options.
- Timezone
  - It can be either local to the user or in a specific timezone.

**Examples**
A lot, like https://www.timeanddate.com/countdown/create

I am familiar with Go, but not with HTML. I have no experience with HTML templating in Go. Maybe I can look into this but not very recently.

Metric-Void:
Added a count timer element that counts the relative time to a specific date.

Closes #427 

image


ErikHacklin:
Hi, any update on this? Looking forward to getting this feature added :)

PR #445, opened 2025-03-16
[RFC] Expose serveApp so we can embed Glance

rubiojr:
I've been toying with the idea of building a small desktop app powered by Glance, fully packaged and without the dependency of an external browser. To do that, Glance would need to expose a public package that I could reuse. 

This is how such a public interface could look like, starting simple.

https://github.com/rubiojr/glance-embed is a sample app embedding Glance using this exposed package.

@svilenmarkov would something like this be accepted?

Mostly a request for comments, to figure out if this would be a welcome addition.
If something like this is to be accepted, it may be a good idea to add a context argument, so the Glance server can be shutdown gracefully programmatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,59 +264,31 @@ Glance can also be installed through the following 3rd party channels:
 
 <br>
 
-## Building from source
-
-Choose one of the following methods:
-
+## Common issues
 <details>
-<summary><strong>Build binary with Go</strong></summary>
-<br>
+<summary><strong>Requests timing out</strong></summary>
 
-Requirements: [Go](https://go.dev/dl/) >= v1.23
+The most common cause of this is when using Pi-Hole, AdGuard Home or other ad-blocking DNS services, which by default have a fairly low rate limit. Depending on the number of widgets you have in a single page, this limit can very easily be exceeded. To fix this, increase the rate limit in the settings of your DNS service.
 
-To build the project for your current OS and architecture, run:
-
-```bash
-go build -o build/glance .
+If using Podman, in some rare cases the timeout can be caused by an unknown issue, in which case it may be resolved by adding the following to the bottom of your `docker-compose.yml` file:
+```yaml
+networks:
+  podman:
+    external: true
 ```
-
-To build for a specific OS and architecture, run:
-
-```bash
-GOOS=linux GOARCH=amd64 go build -o build/glance .
-```
-
-[*click here for a full list of GOOS and GOARCH combinations*](https://go.dev/doc/install/source#:~:text=$GOOS%20and%20$GOARCH)
-
-Alternatively, if you just want to run the app without creating a binary, like when you're testing out changes, you can run:
-
-```bash
-go run .
-```
-<hr>
 </details>
 
 <details>
-<summary><strong>Build project and Docker image with Docker</strong></summary>
-<br>
+<summary><strong>Broken layout for markets, bookmarks or other widgets</strong></summary>
 
-Requirements: [Docker](https://docs.docker.com/engine/install/)
+This is almost always caused by the browser extension Dark Reader. To fix this, disable dark mode for the domain where Glance is hosted.
+</details>
 
-To build the project and image using just Docker, run:
+<details>
+<summary><strong>cannot unmarshal !!map into []glance.page</strong></summary>
 
-*(replace `owner` with your name or organization)*
+The most common cause of this is having a `pages` key in your `glance.yml` and then also having a `pages` key inside one of your included pages. To fix this, remove the `pages` key from the top of your included pages.
 
-```bash
-docker build -t owner/glance:latest .
-```
-
-If you wish to push the image to a registry (by default Docker Hub), run:
-
-```bash
-docker push owner/glance:latest
-```
-
-<hr>
 </details>
 
 <br>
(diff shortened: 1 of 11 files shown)


svilenmarkov:
I think the changes made in the PR are fairly safe to proceed with, however with the ever-increasing scope of what Glance does, it means that I have to be more mindful when potentially making changes to how the app gets served since this is a very different use case. 

I'm not sure what you plan on doing with the desktop app but I can see the current API of only providing a path to a file not being ideal, as down the line you may want to directly provide the contents of the file through the API. Having the ability to do that would likely require a bunch of changes.

The point about adding a context argument would be a good addition so it makes sense to have that in place before going through with this.

rubiojr:
> however with the ever-increasing scope of what Glance does, it means that I have to be more mindful when potentially making changes to how the app gets served since this is a very different use case.

Indeed, certainly not to be taken lightly for a busy project like this. If it helps you with the decision making, vendoring Glance with a small patch is totally doable on my side.

> I'm not sure what you plan on doing with the desktop app 

It's going to be (at least for now) a private team dashboard, and the only goal with this is to ship it as a single binary that will simplify packaging, distributing and running. I also have to bundle an additional http server that powers a few Glance extension widgets defined in the config, so bundling everything together will simplify things.

> but I can see the current API of only providing a path to a file not being ideal, as down the line you may want to directly provide the contents of the file through the API. Having the ability to do that would likely require a bunch of changes.

For my specific case, it's actually not too bad, since (I think, strategy still needs be tested) I can embed the config in the binary, drop it to /tmp or another private config area and boot pointing at it. Agreed that from a Glance PoV, it's a bit limited and not ideal as a public API you're commiting to.

> The point about adding a context argument would be a good addition so it makes sense to have that in place before going through with this.

Totally. Let me know if you wanna move forward with a public API and I'll have a look at it. No rush :pray:.

svilenmarkov:
> If it helps you with the decision making, vendoring Glance with a small patch is totally doable on my side.

Since you bring this up as an option, would you be okay with doing that for a while until you've gotten a little further into your project's development? That would give you a better idea of the kinds of needs you may have from an API like this. 

> I also have to bundle an additional http server that powers a few Glance extension widgets defined in the config, so bundling everything together will simplify things.

Yeah, I get that. It's why I added the `custom-api` widget which removes the need for additional servers in most cases, but that comes with its own set of limitations. I really wanted to make the extensions dynamically loaded modules which would've allowed you to hook into a lot more things, but [Go's plugins](https://pkg.go.dev/plugin) come with a ton of quirks. I am still in search for something more convenient and powerful, and I'm really hoping that [Lua](https://github.com/Shopify/go-lua) isn't the answer.

rubiojr:
> Since you bring this up as an option, would you be okay with doing that for a while until you've gotten a little further into your project's development? That would give you a better idea of the kinds of needs you may have from an API like this. 

Totally, I'll report back.

> 
> Yeah, I get that. It's why I added the `custom-api` widget which removes the need for additional servers in most cases, but that comes with its own set of limitations. I really wanted to make the extensions dynamically loaded modules which would've allowed you to hook into a lot more things, but [Go's plugins](https://pkg.go.dev/plugin) come with a ton of quirks. I am still in search for something more convenient and powerful, and I'm really hoping that [Lua](https://github.com/Shopify/go-lua) isn't the answer.

I still have to test how custom-api would fit in my case, maybe good enough for some of the widgets. I wasn't available when I started experimenting with Glance.

Re: Go plugins, Any thoughts about using something like [Extism](https://github.com/extism/go-sdk)?

rubiojr:
Closing this for now, I'll post back once I have something more concrete.

rubiojr:
@svilenmarkov I ended up with a simple interface that only required to make 3 internal functions public:

* The public package: https://github.com/rubiojr/gapp/blob/main/pkg/glance/glance.go
* How it's used: https://github.com/rubiojr/gapp?tab=readme-ov-file#examples
* A toy app I'm building: https://github.com/rubiojr/hashup-app/blob/main/docs/screenshot.png
* Some notes I took while creating the package: https://github.com/rubiojr/gapp/blob/main/notes.md

Generally speaking, that simple interface is working quite well, it allows me to:
* Read the config file from anywhere I want (embedded in the binary in this case)
* Define listening address (so I can make it local only and random)
* Easily expose HTTP endpoints from the same binary/process that can be consumed later by custom-api, iframe or extension widgets defined in the embedded config (https://github.com/rubiojr/hashup-app/blob/main/glance.yml)
* Easily package all this in a [macOS .app bundle](https://github.com/rubiojr/hashup-app/blob/main/script/package) or an Linux AppImage for distribution

Let me know what you think!

svilenmarkov:
@rubiojr I'll admit, I had something different in mind when you initially brought up the idea of having Glance inside of a desktop application. 

In HashUp, because `glance.yml` is embedded, there wouldn't be any customization from the user and it's mostly made up of custom widgets, so in a way you're utilizing Glance as a UI framework. I'm very curious what specific problems Glance solves for you in this case since having to build a UI around iframes, HTML inside of YAML, and a bunch of HTTP servers to get data across is quickly going to get unwieldy. 

As for the API, I think those 3 functions should probably be put behind a single public one, as I can see them changing in the future during a refactor or if storage ends up getting added. Something like:

```go
err := glance.ServeAppWithContext(ctx, &glance.AppOptions{
  Host: "localhost",
  Port: 12345,
  Config: []byte("")
})
```

rubiojr:
> In HashUp, because glance.yml is embedded, there wouldn't be any customization from the user. And it's mostly made up of custom widgets, so in a way you're utilizing Glance as a UI framework.

Indeed. I understand this may be a bit limiting for a seasoned front-end dev, but it's working out wonders for me :smile:.

>  I'm very curious what specific problems Glance solves for you in this case since having to build a UI around iframes, HTML inside of YAML, and a bunch of HTTP servers to get data across is quickly going to get unwieldy.

In my particular case, quick dashboard prototyping for apps with the ability to plug-in custom backend for new widgets (thanks to extenstion, custom-api and iframe). I write backend toy apps regularly, that may never see the light (home network apps) or be maintained long term, so being able to write throwaway but useful UIs with some YAML saves me precious time :heart:.

> As for the API, I think those 3 functions should probably be put behind a single public one, as I can see them changing in the future during a refactor or if storage ends up getting added. Something like:

```Go
err := glance.ServeAppWithContext(ctx, &glance.AppOptions{
  Host: "localhost",
  Port: 12345,
  Config: []byte // or io.Reader
})
```

Yep, that's what I did in https://github.com/rubiojr/gapp/blob/main/pkg/glance/glance.go. There's only one public `ServeApp` function in that package, so I can easily update the vendored glance without affecting downstream.

Issue #452, opened 2025-03-17
Docker Container YAML Configuration Support


mike391:
### Description

Instead of using docker labels to configure how containers show up, it would be convenient for a lot of users if we could just use `glance.yml` to configure how containers show up in the `docker-containers` widget.  Some users have many containers and don't want to change all their docker compose files to add labels as it would be time consuming and it might even be a reason they dont give using glance a chance.

The proposed behavior will add a check to see if a container name exists in the `glance.yml` and if it does then use the defined overrides there. If a container is not reverenced in `glance.yml` then the dashboard will continue as usual using docker labels to configure containers. This way users dont have to choose between either configuration option and existing users with labels can take advantage of this as well.  

The proposed `glance.yml` configuration will look like this:

```
          - type: docker-containers
            hide-by-default: false
            containers:
              glance:  # This is the actual container name
                title: "test-glances"
                description: "test-description"
                url: "127.0.0.1:3011/test"
                icon: "si:jellyfin"
```

I will implement this myself.

Issue #334, opened 2025-02-09
Allow Docker containers widget to pull data from proxied socket 


retmas-gh:
Hi.
I liked the sound of a widget showing Docker containers status, however there is one thing that will keep me from that - it seems to allow only the mentioned connection via a socket.

Now, as a security practice, all containers I use that need a connection to Docker just to pull some information (like traefik for example) go through an instance of [docker-socket-proxy](https://github.com/Tecnativa/docker-socket-proxy), which uses `haproxy` to proxy the requests to the Docker socket mounted there, allowing only selected methods/endpoints, specified using predefinied environment variables, itself exposing nothing else than the proxy on port 2375.

The way you would set it up is, you run a separate container with the above alongside the one that needs the Docker information with `CONTAINERS=1` environment variable to allow the `/containers` endpoint and in the container of the application that needs the info you set `DOCKER_HOST=tcp://<container_name>:2375` environment variable, for example:
```
services:
  glance:
    image: glanceapp/glance
    volumes:
      - /data/glance:/app/config
    ports:
      - 127.0.0.1:8080:8080
    restart: unless-stopped
    environment:
      - DOCKER_HOST=tcp://glance-socket-proxy:2375
  socket-proxy:
    image: ghcr.io/tecnativa/docker-socket-proxy
    container_name: glance-socket-proxy
    restart: always
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    environment:
      - CONTAINERS=1
```

With a setup like this, the application requesting the data is only allowed requests that are `GET`/`HEAD` to `/containers`,`/events`,`/ping` and `/version`, preventing it the possibility to start, stop, create, exec, delete or any other action that should probably not be allowed to an application that does not handle such actions anyway.

Of course, with the idea of allowing more hosts, instead of a global variable an option in widget could be used.

As far as I can see, the original commiter, [based on the `docs/configuration.md` file](https://github.com/glanceapp/glance/pull/261/commits/2618346a326d309c881774d04f81afd360440dfa), intented for it to work that way, but the concept changed somewhere between his rework and the release.

Kind regards.

svilenmarkov:
Hey, thanks for providing such a detailed explanation! I think this is a reasonable request and shouldn't be difficult to add, here's what I have in mind config-wise:

```yaml
- type: docker-containers
  providers:
    - type: proxy
      host: tcp://glance-socket-proxy:2375
    - type: proxy
      host: tcp://some-other-server:2375
```

[...]

PR #453, opened 2025-03-17
Feat: Added support for configuring Docker containers through glance.yaml instead of labels

mike391:
To address this feature:
https://github.com/glanceapp/glance/issues/452

Before:
oldview

After:
newview

Sample `glance.yml` used during testing:

```
pages:
  - name: Startpage
    width: slim
    hide-desktop-navigation: true
    center-vertically: true
    columns:
      - size: full
        widgets:
          - type: search
            autofocus: true

          - type: docker-containers
            hide-by-default: false
            readable-names: true
            containers:
              glance:  # This is the actual container name
                title: "Test Glances Name"
                description: "test-description"
                url: "127.0.0.1:3011/test"
                icon: "si:jellyfin"
                hide: false
```

Where an existing docker container with the name `glance` is running.


svilenmarkov:
Hey, thanks for contributing!

This is something I want to add support for, however before doing that I think we need to have a discussion around how it would work in conjunction with [this](https://github.com/glanceapp/glance/issues/334) feature, since that does mean that you may end up with multiple containers having the same name.

uykukacinca:
i like arrays more
```yaml
containers:
  - name: glance # This is the actual container name
    provider: glance-socket
    title: "Test Glances Name"
    ...
```

svilenmarkov (approved):
I'm dropping the plan to add merging container data from multiple sources for now, as it makes things much more complicated and confusing. I've tweaked the implementation a bunch to simply merge the labels upon fetching, which removes the need to do multiple checks later on. Thanks again for contributing!

Issue #305, opened 2025-01-08
[Feature Request] Custom repositories for Releases widget


fyr77:
Hello, thanks for the amazing software first of all!

I was wondering if it would be possible to allow for custom Git instances to be used for the releases widget. I run an instance of Akkoma, which is hosted on the [developer's own](https://akkoma.dev/AkkomaGang/akkoma/releases) Forgejo (fork of Gitea) instance. Maybe this could be done by specifying the software and domain as an optional argument?

KevinFumbles:
I was thinking the same thing, both for Releases and Repository. I host my own Gitea instance and would like to specify the domain to connect to. I'm willing to implement this myself as long as someone isn't already working on it - since it doesn't have a label I assume not.

PR #459, opened 2025-03-17
feat: Add custom url for releases widgets

weriomat:
This PR changes the releases widget by adding a custom url parameter, closes [#305](https://github.com/glanceapp/glance/issues/305).
Please note that currently only custom urls are allowed for the `codeberg` and the `gitlab` forges. 

As this is my first time writing go, I would appreciate any feedback.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1566,6 +1566,14 @@ repositories:
   - codeberg:redict/redict
 ```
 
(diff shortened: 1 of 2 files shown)


Felitendo:
Could you please add support for self-hosed Gitea instances too?

weriomat:
I would add a PR, when this lands. In the meantime you can take a look [here](https://github.com/weriomat/glance/tree/gitea-support).

PR #460, opened 2025-03-17
feat: Tautulli widget

titembaatar:
Adds a Tautulli widget which monitor what is being played on your Plex Media Server.
There is even a progress bar ! Which is static, but it looks cool.

tautulli-widget-preview

Edit: I might do a similar one for Jellyfin as well when I have time.


broizter:
> Edit: I might do a similar one for Jellyfin as well when I have time.

I would LOVE to see a similar widget for Jellyfin!

PR #462, opened 2025-03-17
feat: Use conditional requests for RSS feeds

### LINKED ISSUE #413: Add conditional requests to RSS feed reader
### Description

All requests for RSS feeds are currently unconditional, which means that every time the app is loaded (and the widget cache is stale), every RSS feed is fully reloaded.

This isn't efficient, as many RSS feeds will include `Last-Modified` and `Etag` headers to indicate when a feed was last updated. If a feed hasn't been updated since last fetch, Glance should use a cached version of the feed and avoid downloading the (potentially quite heavy) full feed all over again.

I was turned onto the issue by rachelbythebay, who wrote up a decent review of how "well-behaved" feed readers should make requests:

http://rachelbythebay.com/w/2023/01/18/http/

I'm happy to put together a PR implementing these conditional requests for the RSS reader widget.

charlesharries:
Resolves #413 by caching feed responses along with `Last-Modified` and `Etag` headers for use in successive feed requests (with `If-Modified-Since` and `If-None-Match`.

A bit more context about expected RSS feed request headers can be found here:
https://rachelbythebay.com/fs/help.html

I've also added a `User-Agent` header indicating that the request has come from Glance (with version number) instead of the generic `Go-http-client/1.1`.


svilenmarkov:
Hey, thanks for contributing! 

I'll get this merged for v0.8.

PR #466, opened 2025-03-18
make title-url of pihole and pihole-v6 default to /admin

ralphocdol:
As the title says, PiHole's title-url defaults to `/` which you'd just get error 403

error 403 pihole

This just adds `/admin` to it, not sure about other services though so I only added one for PiHole


svilenmarkov (approved):
Thanks for contributing!

Issue #128, opened 2024-06-04
[Request] Swipe function on the smartphone


DickenSerm:
I think it would be really cool if you could switch between columns in the mobile view by swiping left and right. 

Would that be possible without a lot of effort?  

I really like the dashboard, I'm excited to see what new widgets will appear in the future :)

svilenmarkov:
Hey, thanks for the idea!

I think it would be pretty nice and intuitive being able to swipe between columns. In terms of implementation, I'm not entirely sure how difficult it would be as I haven't played around with any of the `touch*` events so far and something tells me there will be edge cases I'll need to account for. I have a bunch of things planned for the next release so I'll leave this one for v0.7.0

dhanadhan:
It could be helpful to implement swipe gestures for navigating between feeds within a group. For example, if I have five subreddits displayed on a single tile, allowing left and right swipes to cycle through them would be convenient. Currently, it is difficult to navigate/click from the top of the screen. Perhaps swiping in the middle of the screen could move between groups, while swiping near the bottom could switch between columns.

jasii:
> It could be helpful to implement swipe gestures for navigating between feeds within a group. For example, if I have five subreddits displayed on a single tile, allowing left and right swipes to cycle through them would be convenient. Currently, it is difficult to navigate/click from the top of the screen. Perhaps swiping in the middle of the screen could move between groups, while swiping near the bottom could switch between columns.

Yes, I was just thinking this today after I added a bunch of RSS feeds in groups to my page. I view Glance on my phone multiple times a day and it would be really nice to swipe between them. Right now I currently have to swipe the title over and then select the new title tab. It's tedious once you have too many titles to fit horizontally on your phone.

PR #471, opened 2025-03-19
feat: Added simple swipe right/left

Nionor:
#128 
Quick and easy solution to add swipe functionality for smartphones to change column.
The **swipeThreshold** maybe should be changeable from some setting but seems to work quite well overall as is.

### DIFF
--- a/docs/custom-api.md
+++ b/docs/custom-api.md
@@ -238,7 +238,7 @@ Output:
 <div>90</div>
 ```
 
-Other operations include `add`, `mul`, and `div`.
+Other operations include `add`, `mul`, `div` and `mod`.
 
 <hr>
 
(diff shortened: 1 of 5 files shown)


internal/glance/static/js/main.js
```diff
@@ -653,6 +653,36 @@ function setupTruncatedElementTitles() {
...
+        touchstartY = event.changedTouches[0].screenY;
+    }, false);
+
+    document.addEventListener('touchend', function (event) {
+        touchendX = event.changedTouches[0].screenX;
+        touchendY = event.changedTouches[0].screenY;
+
+        const swipeThreshold = 50;
+
+        if (Math.abs(touchendX - touchstartX) > Math.abs(touchendY - touchstartY) && Math.abs(touchendX - touchstartX) > swipeThreshold) {
```
Muxutruk2, line 671: Overall good code but I think this if statement allows angles near 90º to trigger it, I suggest multiplying Math.abs(touchendY - touchstartY) by 2.

Also refactoring those 2 checks into variables is also a good Idea

Muxutruk2 (approved):
I approve of this

internal/glance/static/js/main.js
Nionor: I never played around with touch stuff before so I might be wrong how they work but my thinking was that since the length of x must be greater then y the angle has to be less then 45º which I thought was ok.

Muxutruk2: I understand, the thing is that the code will trigger with a long swipe whose start and endpoints are in a cone of 90° from eachother. 60° makes more sense and maybe a time limit

Nionor: Ok I think I understand what you mean, we just have different ways to look at triangles 😃 But I added the Y*2 so we get a shallower angle.

Nionor:
Ok I added some more checks for when to trigger the left and right swipes, hope that helps.
@ralphocdol that seems like a good idea to add too 👍

Muxutruk2:
This should be merged

hchavez8:
I'm new-ish to Github, and very new to app building. It's interesting to see what factors go into a feature that seems so easy. I would never have thought that angles and timing would be a part of it.

PR #480, opened 2025-03-20
Add Kagi and Startpage as search engines

stripedew:
Added two more search engines:

- Kagi `https://kagi.com/search?q={QUERY}`
- Startpage `https://www.startpage.com/search?q={QUERY}`


svilenmarkov (approved):
Thanks for contributing!

PR #481, opened 2025-03-20
fix: invalid url on footer

MisileLab:
I didn't test, but I think it will work.

### DIFF
--- a/internal/glance/templates/page.html
+++ b/internal/glance/templates/page.html
@@ -72,7 +72,7 @@
     <footer class="footer flex items-center flex-column">
     {{ if eq "" .App.Config.Branding.CustomFooter }}
         <div>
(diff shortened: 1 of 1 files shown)


svilenmarkov:
Hey, this shouldn't be an issue since `.App.Version` is the [git tag of the build](https://github.com/glanceapp/glance/tags), which always includes the `v` in the beginning. Which version/os/arch of Glance is this happening for?

MisileLab:
I'm using nixos, does version requires v${version} instead of ${version}, because nixpkgs set build version to ${version}.

ref: https://github.com/NixOS/nixpkgs/blob/d9394d28ba4b1a773d50987b90f664b80c7fd202/pkgs/by-name/gl/glance/package.nix#L25

svilenmarkov:
Ah, I see. I believe this is something that should be changed in the nix package to include the `v` like it does [here](https://github.com/NixOS/nixpkgs/blob/d9394d28ba4b1a773d50987b90f664b80c7fd202/pkgs/by-name/gl/glance/package.nix#L39), since we use the tag and not the bare version as seen [here](https://github.com/glanceapp/glance/blob/1512718d416f4580f7e7bb60c8d5878320e1f032/.goreleaser.yaml#L24).

MisileLab:
thanks

PR #483, opened 2025-03-20
feat: add fallback icon and image on RSS feeds

ewencodes:
Hi, just a little enhancement on the RSS widget.

This will allow widget users to define a fallback image to quickly identify RSS feed item if not image is provided by the RSS feed provider.

As an example, [CERT-FR](https://www.cert.ssi.gouv.fr/) [RSS](https://www.cert.ssi.gouv.fr/feed/) feed doesn't provide an image.

Before : 
image

After :
image

Thanks a lot for this amazing tool.


ewencodes:
@svilenmarkov did you know if this can be implemented ? it's a simple way to identify rss feed items where there isn't images.
thx in advance

PR #486, opened 2025-03-21
Make clickable area for docker-containers widget items bigger

daot:
Made the clickable area for the docker-containers widget take up more of the width and height of each item. Before, having to click on the text itself did not feel great to use on mobile. This change improves the navigation experience significantly while still allowing the hoverable information to be accessed.


svilenmarkov (approved):
Hey, thanks for contributing! Rather than making the full "tile" clickable which breaks the popover functionality on mobile, I've gone ahead and made the link hitbox take up the full width instead. Here's a before and after of that:

image

This, along with slightly increasing the scale of everything mobile, which will go live in a future release should make this much easier to tap on.

PR #491, opened 2025-03-22
Adding custom config for apple-touch-icon and manifest.json [mobile]

dvdpearson:
I couldn’t find this configuration option documented or mentioned on the roadmap, so I went ahead and took a first pass at adding it. This change aims to give users more control over the dashboard icon when saving it to the home screen on Apple devices. Right now, there’s no way to customize the icon, and this update introduces a way to do that.

### DIFF
--- a/internal/glance/config-fields.go
+++ b/internal/glance/config-fields.go
@@ -3,6 +3,7 @@ package glance
 import (
 	"crypto/tls"
 	"fmt"
+	"html/template"
 	"net/http"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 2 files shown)


dvdpearson:
@svilenmarkov are there any steps I need to take to surface this PR?

svilenmarkov:
Hey @dvdpearson, I think it would be more user-friendly if rather than having to provide your own manifest file, you could edit the individual properties of it from within the config. So, the name of the app and icon source could be config options, the rest would be specified by Glance. We can do this by generating the contents of the manifest on application creation and then using it by defining a route in a similar way to how I've done the CSS bundle on the dev branch. https://github.com/glanceapp/glance/blob/1cf4f520f894b7c81329b79c3c966afaa1785728/internal/glance/glance.go#L254

dvdpearson:
@svilenmarkov oh, that sounds better indeed. I'll try to find time in the next few days to update the PR to reflect these changes. Thanks for your input.

PR #494, opened 2025-03-22
Add filter on already seen links for RSS feeds

HtFilia:
When multiple rss feeds point to the same articles (link) they are displayed multiple times in the widgets. This basic implementation keeps track of the already seen links and will filter out links that have already been seen


svilenmarkov (approved):
Thanks for contributing!

Issue #403, opened 2025-03-07
Add container support for Docker Swarm environments


j0nathanr:
### Description

Currently, Glance uses the local docker socket to gather container information on the system it's running on. In a docker swarm environment, containers are distributed across multiple hosts (nodes) as docker services. As a result, Glance does not display all labeled docker containers running in the swarm, only the containers currently running on the same node as it is.

The glance agent could be used and deployed globally in the swarm too send container information to the Glance app. This can work the same way services like Portainer and Swarmpit gather container information using their agents. From [Portainer's GitHub](https://github.com/portainer/agent) page: 

> The Portainer agent is basically a cluster of Docker API proxies. Deployed inside a Swarm cluster on each node, it allows the redirection (proxy) of a Docker API request on any specific node as well as the aggregration of the response of multiple nodes.

I see no potential downsides as swarm mode can be enabled via an environmental variable which defaults to false. Anyone outside a docker swarm environment can simply omit any swarm related configurations.

mortonpd:
For Docker swarms, would a widget that uses a method closer to [traefiks swarm mode](https://doc.traefik.io/traefik/providers/swarm/) not be more appropriate where labels are found on the services rather than on the containers.

This removes the need for agents - but does mean that glance would need to be deployed to a 'manager' node to access the swarm API.

j0nathanr:
That does seem more appropriate considering the current implementation of docker container support also requires labels. The need to run it on a manager node is fine since the agent route would run on every swarm node anyhow, this just means less overhead.

kuhree:
I've managed to get something working on this

kuhree:
Hey 👋 , I also wanted to see this. Just put up a [PR](https://github.com/glanceapp/glance/pull/517) with a solution I'm using locally. Would love some feedback if it works for you or not.

Here's a [docker image](https://github.com/kuhree/glance/pkgs/container/glance) if you're interested: `ghcr.io/kuhree/glance:pr-517`

Here's how the config looks: 

```
- type: docker-containers
  mode: swarm
```

PR #517, opened 2025-03-25
feat: adds docker swarm support to `docker-containers` widget

kuhree:
I wanted to see docker swarm support (#403) so this is my attempt at that. I ended up just reusing the `docker-containers` widget with a `mode` property to switch between the two.

The config looks like this:
```yml
  - type: docker-containers
    mode: "swarm"
```

The result looks like this: 
(left = mode: "standalone", right = mode: "swarm")
Screenshot 2025-03-25 at 5 22 20 PM

[Docker Container](https://github.com/kuhree/glance/pkgs/container/glance) for testing: `ghcr.io/kuhree/glance:pr-517`

Happy to make changes or adjustments.


mortonpd:
Hello, thanks for working on this.

I am seeing an issue with service status showing shutdown when the service is running and using child/parent services.  

View in glance:
image

View in portainer:
image

mortonpd:
Issue remains after task sorting update.

information from query.sh:

https://pastebin.com/GqntZ7wH

kuhree:
New image available. I definitely agree that the scheduling is the way to go.

PR #542, opened 2025-03-31
Add replaceMatches function to custom-api

ralphocdol:
```html
{{ $description := "I think Ruth's dog1 is cuter than your dog2!" }}
<div><pre>Raw Data:   {{ $description }}</pre></div>
<div><pre>String:     {{ replaceAll "dog" "ferret" $description }}</pre></div>
<div><pre>Regex:      {{ replaceAll "dog\\d" "ferret" $description }}</pre></div>
```
image


svilenmarkov (approved):
Thanks for contributing!

Rather than changing the behavior of the current function which will likely break stuff, I added `replaceMatches` that does this. Also, functions that refer to "matches" will be for regex-related stuff, such as the already existing `findMatch` and `findSubmatch` functions.

PR #543, opened 2025-03-31
Added qBittorrent

ThiemeH:
Screenshot 2025-03-31 at 21 47 15

Heya 👋

Here is a simple qbittorrent widget. Little side note, I think the cache isn't really working, unsure why

There are helper functions in custom-api, is it maybe an idea to make these available globally? Could be useful for more widgets

Example config:
```yaml
  widgets:
    - type: qbittorrent
      url: 
      username: 
      password: 
      limit: 5
```

[qbittorrent rest api documentation](https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-4.1)#get-torrent-list)


shaezzy:
Hey, this is awesome, thanks for adding.

What do you think about making a "torrent widget" and having different clients that can be configured?
I was implementing a similar widget for Deluge #561 and I think we could reuse the front end for different torrent clients.

Let me know what you think!

jdcool00:
why is this not getting pulled it's really something many people would want

PR #558, opened 2025-04-05
Search bar acts like address bar feature implemented

develoopeer:
# Feature from https://github.com/glanceapp/glance/issues/229 issue implemented
That is the second PR i made about this feature. The first [one](https://github.com/glanceapp/glance/pull/256) was closed by me, because i messed up with resolving merge conflicts.

This time all string that looks like link will be interpreted as a valid links. For example:
- `http://google.com`
- `https://google.com`
- `google.com`
- `www.google.com`

Also i want to add a schebang for external links, like it already implemented with youtube for example. Is it a good idea? If it is i will create a separate PR.

### DIFF
--- a/internal/glance/static/js/main.js
+++ b/internal/glance/static/js/main.js
@@ -94,6 +94,27 @@ function updateRelativeTimeForElements(elements)
     }
 }
 
+function isValidUrl(string){
+	let url;
+	try {
+		url = new URL(string);
+	} catch ({ name, message }) {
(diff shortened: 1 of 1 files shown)


develoopeer:
## How schebang works
search

hegde-atri:
Was looking for just this. 

Is there any more work that needs to be done, or are we just waiting for it to be merged?

develoopeer:
I think there is no work that needs to be done in this PR. This  feature originally was planned for v7.0 but was postponed and now is planned for v8.0. Maybe it will be merged just before release.

PR #561, opened 2025-04-06
Add Widget for Deluge Torrent

shaezzy:
Adding a widget for showing the progress of downloading torrents
image

Config
```yaml
- type: torrents
  url: http://localhost:8112/
  username: admin
  password: deluge
  limit: 10
  client: deluge
```

Current implementation includes deluge only, but additional supported torrent clients could be added as well.
The `TorrentClient` interface, can be reimplemented for different torrent clients, and added in the switch statement for `clientType`


jdcool00:
This is such a needed feature imo why it has not been pulled yet any idea?

russssl:
amazing feature

Issue #568, opened 2025-04-08
A way to wait for previous request in custom-api


ralphocdol:
### Description

As the title says.
I've come across few instances where this can be useful
1. from a `What's up Docker` widget in community-widgets https://github.com/glanceapp/community-widgets/blob/main/widgets/wud-monitor-by-panonim/README.md where it would have been useful to have the POST request as a main request and have the GET request be on the subrequest and making it wait for post to be finished.
2. half part from this feature request: https://github.com/glanceapp/glance/issues/563

I'll submit a PR for this

eskadrggj:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

svilenmarkov:
This is now possible in >=v0.8.0.

PR #569, opened 2025-04-08
add wait-last-request for custom-api

ralphocdol:
For https://github.com/glanceapp/glance/issues/568

#### Config example:
```yaml
- type: custom-api
  url: https:${WEBSERVER_URL}/data          # Replaces the value of the key timestamp inside file data.json with the current timestamp
  method: post                              # Notice post
  skip-timeout-error: true                  # Example in the sub2, this should skip timeout error whenever it reaches it as the data from this request isn't needed
  subrequests:
    sub1:
      url: https:${WEBSERVER_URL}/data          # A get request the fetches a file data.json that has a timestamp key
      wait-last-request: true                   # The feature added, this waits for the previous request
    sub2:
      url: https:${WEBSERVER_URL}/slow          # Just a request that finishes up after 10s to simulate timeout
      skip-timeout-error: true                  # Just an example
  cache: 1s
  template: |
    <div><pre>POST:     {{ (.Subrequest "sub1").JSON.String "timestamp" }}</pre></div>
    <div><pre>GET:      {{ .JSON.String "timestamp" }}</pre></div>
    <div><pre>Matched?: {{ eq (.JSON.String "timestamp") ((.Subrequest "sub1").JSON.String "timestamp") }}</pre></div>
    <div><pre>SUB2:     {{ (.Subrequest "sub2").Response.Status }}</pre></div>
```

#### Output:
```
POST:     1744133075172
GET:      1744133075172
Matched?: true
SUB2:     Get "https://<redacted>/slow": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

Setting `wait-last-request: false` or not setting it at all would make the `Matched?:` value randomized (GET would get the previous data before the POST happened or GET the latest POST data) since requests are concurrent and would it be unpredictable.

> [!CAUTION]
>
> Multiple SUCCESSIVE usage of this property in the same widget can slow your page load time depending on the API's response by another 1~5 seconds (timeout limit) each `wait-last-request`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1342,7 +1342,7 @@ Examples:
 | allow-insecure | boolean | no | false |
 | skip-json-validation | boolean | no | false |
 | wait-last-request | boolean | no | false |
-| ignore-timeout | boolean | no | false |
+| skip-timeout-error | boolean | no | false |
 | template | string | yes | |
 | parameters | key (string) & value (string|array) | no | |
 | subrequests | map of requests | no | |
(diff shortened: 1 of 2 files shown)


anant-j:
Hi @svilenmarkov are there any plans to merge this PR in the near future?
Thanks

svilenmarkov:
@anant-j I'll add something that would allow you to achieve the same result in a more imperative way in v0.8.

PR #583, opened 2025-04-14
﻿[sensors] allow sensor readings when there are warnings

anxdpanic:
New to Go, I'm not entirely sure the `nil` checks are required.

Before:
```
/app # ./glance sensors:print
Failed to retrieve list of sensors: Number of warnings: 1
 - read /sys/class/hwmon/hwmon4/temp1_input: no data available
```

After:
```
/app # ./glance sensors:print
Errors encountered while retrieving list of sensors:
 Number of warnings: 1
 - read /sys/class/hwmon/hwmon4/temp1_input: no data available
Sensors found:
 acpitz: 27.8°C
 acpitz: 29.8°C
 pch_skylake: 36.5°C
 jc42: 27.6°C
 coretemp_package_id_0: 59.0°C
 coretemp_core_0: 56.0°C
 coretemp_core_1: 59.0°C
 coretemp_core_2: 59.0°C
 coretemp_core_3: 56.0°C
```
image

### DIFF
--- a/internal/glance/cli.go
+++ b/internal/glance/cli.go
@@ -96,22 +96,22 @@ func parseCliOptions() (*cliOptions, error) {
 func cliSensorsPrint() int {
 	tempSensors, err := sensors.SensorsTemperatures()
 	if err != nil {
-		fmt.Printf("Failed to retrieve list of sensors: %v\n", err)
+		fmt.Printf("Errors encountered while retrieving list of sensors:\n %v\n", err)
 		if warns, ok := err.(*sensors.Warnings); ok {
 			for _, w := range warns.List {
 				fmt.Printf(" - %v\n", w)
 			}
 		}
-		return 1
 	}
 
-	if len(tempSensors) == 0 {
+	if tempSensors == nil || len(tempSensors) == 0 {
 		fmt.Println("No sensors found")
 		return 0
 	}
 
+	fmt.Println("Sensors found:")
 	for _, sensor := range tempSensors {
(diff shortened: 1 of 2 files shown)


svilenmarkov (approved):
> New to Go, I'm not entirely sure the nil checks are required.

In about 99% of cases, if `err` isn't `nil` you're not supposed to use the value since it may result in nil pointer dereference panic down the line or the value may just be a dud with zero values because it couldn't be populated properly. The exception to the rule is cases like this where something may fail in numerous ways, some of which aren't catastrophic, and the value is still useable. 

Maybe my exposure to Go codebases is lacking, but so far I haven't seen a consistent way to handle this scenario and it seems like everyone just makes up their own error types, and the only way to know this is reading the docs or the code. Somewhat frustratingly, gopsutil's docs don't even mention the existence of the `Warnings` type and I only found out about it a few days ago by skimming through the code. It would've taken me a couple more trips to its codebase before discovering what you have, which is that you can use the value even if `err` isn't `nil`, so thank you for that. 😅 

Note however that we should only assume the value is useable if `err` is of type `Warnings` and not otherwise, since there are [situations in which the error applies to all sensors](https://github.com/shirou/gopsutil/blob/3608872740b7ded1691742bcd48fe5354a8eb1a6/sensors/sensors_linux.go#L27), not just individual ones, so I've gone ahead and made that change. 

Btw, doing both checks is unnecessary:

```go
if tempSensors == nil || len(tempSensors) == 0 {
```

since `len()` of `nil` value will safely return 0. 

Thanks for contributing.

PR #591, opened 2025-04-16
fix healthcheck when alternate port provided

anxdpanic:


afro-systems:
Any chance this will be getting merged?

svilenmarkov:
As per the above comment, the health check has now been removed.

PR #594, opened 2025-04-17
update error reporting for custom-api JSON validation

anxdpanic:
show the request error on non-200 responses

Screenshot 2025-04-16 203603

### DIFF
--- a/internal/glance/widget-custom-api.go
+++ b/internal/glance/widget-custom-api.go
@@ -203,13 +203,18 @@ func fetchCustomAPIRequest(ctx context.Context, req *CustomAPIRequest) (*customA
 	body := strings.TrimSpace(string(bodyBytes))
 
 	if !req.SkipJSONValidation && body != "" && !gjson.Valid(body) {
-		truncatedBody, isTruncated := limitStringLength(body, 100)
-		if isTruncated {
-			truncatedBody += "... <truncated>"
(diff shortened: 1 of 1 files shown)


svilenmarkov:
We could maybe do this for status codes >= 500 because their meaning is ubiquitous, though I've purposefully left handling the response status codes to the template, since servers can return any 4xx code to signify any kind of meaning, which you'd likely want to be able to handle manually on a per-widget basis. For example a 404 could mean a "this endpoint doesn't exist, the user provided an incorrect path and needs to fix their URL", or it could mean "this resource doesn't exist yet but that's expected behavior" and the template can conditionally show data from that endpoint or not.

anxdpanic:
I may have read/followed the code incorrectly. 

It seemed that if json validation happened at all. `Invalid response JSON` was the only possible return to the template with no handling. 
Here, it's already been confirmed that the JSON wasn't valid, so it's now  checking if the expectation of it being valid was also true (200 response). This lets the user configuring, see the error response of that validation without going to the logs directly since `Invalid response JSON` was the only possible return previously.

If `skip-json-validation` is enabled the status code and error handling are still in the hands of the template.

svilenmarkov (approved):
Ah, you're right, my bad. The description of the PR made it sound like it was a part of the normal request validation flow so I didn't pay too much attention to the code. This is a good change, thanks for contributing!

PR #600, opened 2025-04-20
Add new layout option to bookmarks

0x00cl:
I never liked the list layout of the bookmarks widget, so I thought of adding a new one. As far as I can tell no one has requested this feature or option but I'd like to see it added.

This sort of "quick links" can be seen in Firefox default home page and Google Chrome and probably other browsers too.
firefox-home

This is what the new layout (left) I added looks like, with also the default one (right).

bookmarks-widget-preview

- To enable the layout you need to add the configuration `layout: grid` to the bookmark widget.
  - If no layout is provided or an option that doesn't exist then it defaults to the old one.
- If the user doesn't provide an Icon then it will use the first character of the `title`
  - If the user doesn't provide a title either then it simply defaults to the "G" character (Glance)

EDIT: Now that I think about it, the icons could be more square... what do you think?


svilenmarkov:
Thanks for contributing, I've added the feature in [this](https://github.com/glanceapp/glance/commit/df9b886bd04ffad08244c08e3eeb59a66a6e28c0) commit.

PR #602, opened 2025-04-20
Dynamize manifest.json with configurable options

dvdpearson:
I've made some exciting updates to my original [PR](https://github.com/glanceapp/glance/pull/491)! It's now more user-friendly and configurable, and you won't need to ask for the manifest file explicitly anymore.

I noticed that this configuration option wasn't documented or mentioned on the roadmap, so I decided to take the initiative and add it. This change is all about giving users more control over the dashboard icon when saving it to the home screen on Apple/PWA devices. Currently, there’s no option to customize the icon, and I'm excited to share that this update will introduce that possibility!

@svilenmarkov what do you think?


svilenmarkov (approved):
Thanks for contributing and making those changes! I've made a bunch of additional tweaks, more importantly parsing and caching the manifest contents only once on application creation rather than parsing on each request, as well as changing how the caching works, since the manifest can change between config reloads, which the `staticFSHash` wouldn't reflect.

PR #623, opened 2025-04-29
Minor design tweak, polishing up the polyline corners 

mmshivesh:
Adds a `stroke-linejoin="round"` to the polyline to ensure that they connect cleanly and look cohesive with the rest of the design of Glance. Ref: [MDN docs for polyline joins](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/stroke-linejoin#round). This shouldn't be noticeable for the vast majority of folks and is usually only visible at higher zoom levels. I've included a screen. Included a few screenshots of the changes:

Dark Mode: 

Light Mode:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -5,6 +5,7 @@
   - [Auto reload](#auto-reload)
   - [Environment variables](#environment-variables)
   - [Including other config files](#including-other-config-files)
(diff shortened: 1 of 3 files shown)


svilenmarkov (approved):
Thanks for contributing!

PR #628, opened 2025-04-30
RTL property for base widget

pypp:
I didn't documnte the change because I don't even know if It's in the scope of the project. so let me know so that I will add documentation accordingly.


svilenmarkov:
Technically you can already do this by adding the following to each widget:

```yaml
class-name: rtl
```

Though I don't know how well that would work in some widgets because none were optimized for it.

PR #662, opened 2025-05-15
Add http proxy support

mazzz1y:
I use Glance in a network-restricted environment, and I need an HTTP proxy to connect to third-party services.  
This change adds support for the http_proxy and https_proxy environment variables in the native Go way.

Before the change:
```
2025/05/15 06:08:23 ERROR Failed to fetch market data symbol=BTC-USD error="Get \"https://query1.finance.yahoo.com/v8/finance/chart/BTC-USD?range=1mo&interval=1d\": dial tcp: lookup query1.finance.yahoo.com on 127.0.0.11:53: server misbehaving"
```


svilenmarkov (approved):
Thanks for contributing!

PR #678, opened 2025-05-18
Fix text-truncate or related formatting

ralphocdol:
This addresses the weird formatting that's happening when using `text-truncate`, `single-line-titles` and similar classes
image
image

Test config:
```yml
- type: html
  title: Playground
  source: |
    <div class="text-truncate">
      I am a sample text for text-truncate that will truncate a long text
    </div>
    <div>-----</div>
    <div class="text-truncate-2-lines">
      I too am a sample text to text text-truncate-2-lines
    </div>
    <div class="text-truncate"></div>
```

Edit: 
I feel like doing `.replace(/\s+/g, ' ').trim()` to also remove new lines before the trim would be better, but maybe that's just me.

### DIFF
--- a/internal/glance/static/js/page.js
+++ b/internal/glance/static/js/page.js
@@ -661,7 +661,7 @@ function setupTruncatedElementTitles() {
 
     for (let i = 0; i < elements.length; i++) {
         const element = elements[i];
-        if (element.getAttribute("title") === null) element.title = element.textContent.trim();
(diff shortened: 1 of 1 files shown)


svilenmarkov (approved):
I changed it to `element.innerText` which resolves both the issues you mentioned, since it automatically strips out all whitespace. Thanks for contributing!

PR #686, opened 2025-05-22
chore(license): update copyright info in LICENSE file

1411430556:
update copyright info in LICENSE file


svilenmarkov:
I'm no lawyer, but the name Glance isn't registered, so I believe it has no effect. I don't mind leaving the license as is, since that's what most projects do.

PR #689, opened 2025-05-22
Correct docs for Go's `text/html` built-in template functions

thallada:
Source: https://pkg.go.dev/text/template#hdr-Functions

The comparison functions the greater-than-or-equal-to and less-than-or-equal-to are `ge` and `le`, not `gte` and `lte`. I got an error rendering my template before I realized this.


svilenmarkov (approved):
Whoops, sorry about that!

PR #693, opened 2025-05-24
build: bump purego to build against newer go

chenrui333:
there is some issue with purego 0.8.2, bump it to the latest 0.8.4

```
  link: duplicated definition of symbol dlopen, from github.com/ebitengine/purego and github.com/ebitengine/purego
```

relates to:

- https://github.com/ebitengine/purego/issues/313


svilenmarkov (approved):
Thanks for contributing!

Issue #700, opened 2025-05-26
Search Suggestions so that Google, DDG, and etc. have search suggestions.


rhijjawi:
### Description

Most search services already provide an API for Search Suggestions, for example:
[https://www.google.com/complete/search?client=chrome&q={QUERY}](https://www.google.com/complete/search?client=chrome&q=example)
```json
[
  "example",
  [
    "example deutsch",
    "examples",
    "https://example.com"
  ],
  [
    "",
    "",
    "",
    "",
    "",
    "",
    "",
    "",
    "",
    "",
    "",
    "",
    "",
    "",
    ""
  ],
  [],
  {
    "google:clientdata": {
      "bpc": false,
      "tlw": false
    },
    "google:suggestrelevance": [
      602,
      601,
      600,
      561,
      560,
      559,
      558,
      557,
      556,
      555,
      554,
      553,
      552,
      551,
      550
    ],
    "google:suggestsubtypes": [
      [
        512
      ],
      [
        512
      ],
      [
        44
      ],
      [
        512
      ],
      [
        512
      ],
      [
        512
      ],
      [
        512
      ],
      [
        512
      ],
      [
        512
      ],
      [
        512
      ],
      [
        512
      ],
      [
        512
      ],
      [
        512
      ],
      [
        512
      ],
      [
        512
      ]
    ],
    "google:suggesttype": [
      "QUERY",
      "QUERY",
      "NAVIGATION",
      "QUERY",
      "QUERY",
      "QUERY",
      "QUERY",
      "QUERY",
      "QUERY",
      "QUERY",
      "QUERY",
      "QUERY",
      "QUERY",
      "QUERY",
      "QUERY"
    ],
    "google:verbatimrelevance": 1300
  }
]
```

rhijjawi:
Would like to clarify I can get started on an implementation for this;

The idea is have a config such as below, and render the results.
```yaml
- type: search
   search-engine: google
   suggestions: true
   bangs:
       - title: YouTube
       shortcut: "!yt"
       suggestions: true
       url: https://www.youtube.com/results?search_query={QUERY} #Since hardcoded URL AND suggestions is true, needs suggestions_url
       suggestions_url: https://www.google.com/complete/search?client=chrome&ds=yt&q={QUERY}
```

PR #702, opened 2025-05-26
Add search suggestions feature with separated search engine and suggestion engine.

rhijjawi:
Implemented a new handleSearchSuggestionsRequest HTTP handler that:
- Accepts a query and a search engine as parameters.
- Fetches search suggestions from predefined engines (google, duckduckgo, bing, startpage).
- Added support for parsing JSON responses from suggestion APIs (Google-style format).
- Extracted helper function createResponse() to parse suggestion responses cleanly.
- Set appropriate HTTP error codes for bad requests and upstream failures.
- Ensured results are returned as structured JSON:
```json
{
  "query": "user input",
  "suggestions": ["suggestion1", "suggestion2", "suggestion3"]
}
```

For example, if 
```yaml
search-engine: google
suggestions: true
```

The search bar will use Google's autocompletion API to generate _autocompletions_ for the query.
In the case that Startpage is the search engine, but `suggestion-engine` is defined, the `suggestion-engine` value takes precedence.

```yaml
search-engine: startpage
suggestions: true
suggestion-engine: google
```

### To be entirely clear:
[./glance.go](https://github.com/rhijjawi/glance/blob/cf5530014ea2f5e37ed8c5d8e5665f9a0638a442/internal/glance/glance.go#L406) now contains the rough equivalent of a proxy. Its sole purpose is to ignore Startpage's CORS restriction on who can make requests to their suggestion endpoint, this is within their Terms of Use.
This seems like the best solution at the moment, it works very well. I have not yet run into issues regarding rate-limiting, nor do I predict any issues.

### DIFF
--- a/internal/glance/glance.go
+++ b/internal/glance/glance.go
@@ -467,9 +467,10 @@ func (a *application) handleSearchSuggestionsRequest(w http.ResponseWriter, r *h
 
 	json.NewEncoder(w).Encode(result)
 }
+
 func createResponse(rawResponse []any, suggestionEngine string) []string {
 	suggestions := []string{}
-
+	// Handle Google/Bing-style responses: [{params : "s1"}, {params : "s2"}, ...]
 	if suggestionEngine == "duckduckgo" {
 		for _, item := range rawResponse {
 			if m, ok := item.(map[string]interface{}); ok {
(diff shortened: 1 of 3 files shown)


rhijjawi:
Issue #700 - Feature Request Implementation

Issue #703, opened 2025-05-26
Multiple todos in a group do not work properly


tty2:
### Description

If you create multiple `to-do` widgets inside a `group`, only first allow you to add a new task. Others do not work.
 
- Installed inside Unraid as an app from store (so it's a docker version).
- Version 0.8.3
- yaml
```- size: full
            widgets:
                - type: group
                  widgets:
                      - type: to-do
                        id: first
                        title: first
                      - type: to-do
                        id: second
                        title: second
```
- Screenshots

first
![Image](https://github.com/user-attachments/assets/2484b514-965a-4825-9e01-5ed2d630fc48)
second

![Image](https://github.com/user-attachments/assets/d31cf73c-baf9-46fd-8130-1699a91c258e)

- As I can see it doesn't depend on browser or OS.
- To reproduce use yaml provided above

rhijjawi:
Fixed in #705.

PR #705, opened 2025-05-27
setupTodos using incorrect logic and skipping to-do item.

rhijjawi:
After looking over #703, I realized that `todo.default(elems[i])` in [page.js](https://github.com/glanceapp/glance/blob/b94647efc9145beb3a47368b30f633cc8e94acc2/internal/glance/static/js/page.js#L652C9-L652C32) was overwriting elements while still reading over them in the loop. This causes a Todo element to "disappear" as it takes the position of the overwritten element and being skipped in the iteration.

The fix is to create a copy of the element array so that it is maintained while changes are being made to the document.

### DIFF
--- a/internal/glance/static/js/page.js
+++ b/internal/glance/static/js/page.js
@@ -643,13 +643,14 @@ async function setupCalendars() {
 }
 
 async function setupTodos() {
-    const elems = document.getElementsByClassName("todo");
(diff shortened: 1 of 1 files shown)


svilenmarkov (approved):
Thanks for contributing!

PR #706, opened 2025-05-28
Add search suggestions (with configurable engine) and configurable shortcuts

kylerm42:
## Summary

  • Enhanced search widget with configurable shortcuts for quick site navigation and fuzzy matching
  • Added search suggestions integration with popular search engines (Google, DuckDuckGo, Bing, Startpage)
  • Navigate directly to URLs, bypassing search engine

### Shortcuts System:
  - Fuzzy matching against shortcut titles and optional alias text
  - Dropdown interface with keyboard navigation (arrow keys, tab, enter)
  - Exact match prioritization for immediate navigation (e.g. type `gm`, then `Enter` to go directly to Gmail)

```
  shortcuts:
    - title: Gmail
      url: https://mail.google.com
      alias: gm
    - title: GitHub
      url: https://github.com
```

### Search Suggestions:
  - Suggestions from configured search engines with debouncing
  - Automatic fallback to main search engine if not specified
  - Support for custom suggestion endpoints
  - New /api/search/suggestions endpoint to handle requests

```
  suggestions: true
  suggestion-engine: duckduckgo  # or google, bing, startpage
```

### Direct URL Navigation:
  - Detects when the query is a URL and takes the user there instead of searching for it
  - Supports standard URLs (`google.com`, `www.google.com`), HTTP(S) prefixes (`https://google.com`), localhost (`localhost:3000`) and IP addresses (`192.168.1.1`), all with optional ports

### Keyboard Navigation:
  - ↓/Tab and ↑/Shift+Tab for dropdown navigation
  - Enter to select highlighted item or perform search
  - Click support for all dropdown items
  - Maintains existing bang functionality and keyboard shortcuts
 
image

This is similar to #702, but with some extra features. Also relates to issue #700

PR #712, opened 2025-06-02
Support regular background widget refresh

kylerm42:
I, like probably many others, am using Glance as a home page and new tab page. The search bar is where I start most of my navigation (hence #706), but I am frequently waiting for widgets to refresh before anything shows up. This PR aims to address some of these issues and ensure page loads are speedy every time.

- Predictive Widget Refresh
  - Updates widgets that are outdated OR will become outdated before the next refresh cycle
  - Prevents widgets from being stale when users visit pages
  - Example: With 15min refresh interval, a widget expiring in 5min gets updated proactively

- Configurable
  - background-refresh-enabled (default: false) - opt-in
  - background-refresh-interval (default: 15m) - customizable refresh frequency

```yaml
  server:
    background-refresh-enabled: true
    background-refresh-interval: 10m
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -257,6 +257,8 @@ Server configuration is done through a top level `server` property. Example:
 server:
   port: 8080
   assets-path: /home/user/glance-assets
+  background-refresh-enabled: true
+  background-refresh-interval: 15m
 ```
 
 ### Properties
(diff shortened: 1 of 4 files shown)


gitnew12:
This is awesome. Another improvement is load widgets dynamically similar to getHomepage.

Load the dashboard instantly but for custom-api widgets, allow for empty `<div>` placeholders while API data is being fetched. Once request is processed, inject data into the placeholder `<div>`s. With the option to refresh the specific custom-api widgets on an interval.

Issue #643, opened 2025-05-07
Navbar Shortcut for Desktop


cartwatson:
### Description

`ctrl+shift+number` to navigate between headers

- What problem would this feature solve?
  - Navigating multiple pages becomes easier and quicker
- Are there any potential downsides to this feature?
  - None that I can think of, if a user doesn't want to use them they don't have to and I chose `ctrl+shift+number` as that shouldn't clobber any system level shortcuts (these are used in the outlook webapp)
- If applicable, what would the configuration for this feature look like?
  - I don't see a need for this to be configurable but I could see how someone else might. If config was desired it would need to be a glance-level config option(s), but I think that it's more mess than it's worth from an implementation and user side.
  - Could add config to enable/disable the setup function
- Are there any existing examples of this feature in other software?
  - As previously mentioned this is used in the web version of Outlook to navigate between inbox, calendar, etc
- If applicable, include any external documentation required to implement this feature
  - Only needs a line to explain the shortcut, adding config would mean more docs

I'd like to implement this and have a rough draft in [cartwatson/glance-fork](https://github.com/cartwatson/glance-fork/tree/jwatson/navbar-shortcuts)

PR #726, opened 2025-06-11
feat: navbar shortcuts

cartwatson:
Reference https://github.com/glanceapp/glance/issues/643.
Add navbar shortcuts with `ctrl+shift+#` where # is the tab from left to right. This only works for up to ten tabs, if desired I can change to be better in line with most browsers and have digit 9 go to the last tab instead of the ninth. Made the assumption that most users don't have 9+ tabs/pages


ralphocdol:
Although this is a good feature but

> Are there any potential downsides to this feature?
>    None that I can think of, if a user doesn't want to use them they don't have to and I chose ctrl+shift+number as that shouldn't clobber any system level shortcuts

I don't see an option for it to be disabled, I think there should be one since Linux DE or WM etc. have customizable shortcuts that could conflict with this.

PR #734, opened 2025-06-17
Fix DNS summary tap highlight

ralphocdol:
Chrome based mobile view with the summary tag has a tap highlight that does not cover the entire area.
Screenshot_20250617_185913

<details>
<summary>Personally, I'd apply this to a system level.</summary>

```css
*:active {
    -webkit-tap-highlight-color: transparent;
}
```
</details>

### DIFF
--- a/internal/glance/static/css/site.css
+++ b/internal/glance/static/css/site.css
@@ -61,6 +61,10 @@ button {
     width: 10px;
 }
 
+*:active {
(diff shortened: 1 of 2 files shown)


svilenmarkov (approved):
Thanks for contributing! I went ahead and made it global since I did notice it in a couple of other places and it looked off.

PR #746, opened 2025-06-23
Added Object helper for custom-api

ralphocdol:
There are API's that list items as Objects instead of Array similar to this discussion https://github.com/glanceapp/glance/discussions/742, Mostly RPC endpoints

This adds a helper function `Object` a bit similar to `Array` except this has `.Key` and `.Value` helper itself.

Sample API data:
```json
{
    "object": {
      "key1": "value1",
      "key2": "value2",
      "key3": {
        "test1": "v1",
        "test2": "v2"
      },
    }
}
```

Config:
```yml
- type: custom-api
  url: http://webserver.local/sample-object
  cache: 1s
  template: |
    <div>Objects</div>
    {{ range .JSON.Object "object" }}
      <div>{{ .Key }}: {{ .Value.String "" }} | {{ .Value.String "test1" }}</div>
    {{ end }}
    <br>
    <div>Length: {{ len (.JSON.Object "object") }}</div>
```

Result:
Screenshot_20250623_204314

<sub>*Unless there's already a way? couldn't find one.*</sub>

### DIFF
--- a/docs/custom-api.md
+++ b/docs/custom-api.md
@@ -445,10 +445,8 @@ The following functions are available on the `JSON` object:
 - `Float(key string) float`: Returns the value of the key as a float.
 - `Bool(key string) bool`: Returns the value of the key as a boolean.
 - `Array(key string) []JSON`: Returns the value of the key as an array of `JSON` objects.
-- `Object(key string)`: Returns the value of the key as a list of properties.
-  - `.Key`: the property name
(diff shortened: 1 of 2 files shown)


ralphocdol:
> Let me know if you have any change suggestions, otherwise I'm happy to get it merged like this.

I am good with this. So there really was a way already.

PR #752, opened 2025-06-28
Add sort-by option for videos widget

igorkim:
Added a sort-by option for videos to control their ordering. For example, in my “Watch Later” playlist, I can now sort videos by the date they were added.


svilenmarkov (approved):
Thanks for contributing!

PR #763, opened 2025-07-15
Removed "from_version" url parameter for the changedetection.io widget.

juvanj:
The changedetection.io widget was not working anymore. The from_version url parameter was passed with the id of the current watch -1. Since changedetection.io is not counting sequentially it was not working in my installation. So there I removed the useless parameter (it compares anyways to the previous watch).

I would love to see my change in the main branch of this project.

Also very thanks to all the people contributing.


juvanj:
Please someone to check my changes?

PR #784, opened 2025-08-03
Fix vertically scrollable nav

ralphocdol:
Navigation header can be vertically scrolled.
Screencast_20250803_172033


svilenmarkov (approved):
Thanks for contributing! 

Took me a little while to figure out why this was happening and it turns out it was because of the fixed `line-height` on each `.nav-item` which wasn't accounting for the bottom border. Instead of relying on that to center the text, I just added an extra element and centered that, which should be a bit more robust in the event that more things get added there.

PR #797, opened 2025-08-14
Thumbnail support for the RSS style vertical-list

ralphocdol:
```yaml
- type: rss
  # style: vertical-list
  show-thumbnails: true
```

image


michaelwschultz:
@ralphocdol you're killing it with these changes. Appreciate the contributions 🥂

Issue #800, opened 2025-08-17
Ability to specify HTTP Method (GET, POST, ...) when using "newRequest" template function.


MartyX98:
### Description

custom-api widget's go template function "newRequest" has sub functions "withHeader" and "withParameter" but has no function to specify the HTTP Method request type. I would like to add a function "withMethod" to specify the method type. The idea is to be able to do this;

```
{{ 
  $response := newRequest ${API_URL}
    | withMethod "POST"
    | withParameter "paramName" "paramValue"
    | getResponse
}}
```

As per the contribution's guidelines, I am making this feature request first and I intend to develop this myself. I will link the PR both here and this feature request in the PR.

samcro1967:
For anyone following this request: this functionality has been implemented and tested in the samcro1967/glance fork in PR #79. The implementation adds `withMethod` for dynamically created Custom API requests while preserving the existing GET default and body-implies-POST behavior.

Fork implementation: https://github.com/samcro1967/glance/pull/79

PR #801, opened 2025-08-17
Addition of "withMethod" func to specify http method request type after using "newRequest" func

MartyX98:
I have added function "withMethod" to be able to specify the HTTP method request type after using "newRequest" method. This is useful for when you need to make a request with a method type different than the default "GET".

I have tested the implementation by querying my Pihole's API endpoint using "GET", "POST", and "DELETE" method types and all went successfully or as expected.

Resolves: #800


PapaZ810:
@svilenmarkov can you merge dev to main to make a release out of this. Working on a widget and I believe this is the final issue preventing it from working. Thanks!

samcro1967:
A data point for this implementation: `withMethod` has been incorporated and tested in the samcro1967/glance fork. It preserves GET as the default and the existing body-implies-POST behavior while allowing an explicitly selected method for dynamically created requests.

Fork integration:
https://github.com/samcro1967/glance/pull/79

PR #803, opened 2025-08-19
Todo widget: use a persistent SQLite database instead of the browser's local storage

MarioKuka:
1. _What is the motivation behind this change?_ 
It's simple: Todo list that works with the browser's local storage is practically useless to me and makes no sense...

2. _Is this code generated by AI?_ 
Yes, absolutely all of it.

3. _So is this code a piece of crap?_ 
Maybe yes, maybe no, but it works very well and I currently have it deployed on my server.

If it makes sense to you to have this change in the master branch, let me know. It needs a little love to make it production-ready.


MarioKuka:
> I agree with this suggestion. Just a question: would I then be able to access this same SQLite database by another container? My idea is in docker compose to create a mapping of this file to the outside and then map the second container to access this file and do stuff with it, something like:
> 
> ```
> volumes:
>  - /glance/sqlite.db:/home/user/sqlite.db
> ```

This should already be possible in this version. For each Todo widget, you can use the `data-path` parameter to specify the directory where the database is located, the default value is `./data` and multiple todo lists can use same database. (From the root perspective in the Docker container, the full path will be `/app/data/`). 

Te name of the database is not configurable and must be named as `todos.db`  so the full path inside the docker container is `/app/data/todos.db`

For example, if you have this Todo list:  
```
- type: to-do
  title: Personal Tasks
  id: personal
  data-path: ./data # 
```
and you want your database for this to-do list to be accessible from outside, it should be enough to add:
```
volumes:
 - /glance/todos.db:/app/data/todos.db
```

Yes, it's a bit complicated and needs to be simplified.
Requiring the exact name of the database todos.db is not ideal.

WAY29:
go.mod and go.sum update is required, this PR imoprt a new dep: modernc.org/sqlite

MarioKuka:
Much better version here #838

PR #808, opened 2025-08-28
feat: Listen on Unix Socket

AWildLeon:
This pull request adds support for running the server on a Unix socket in addition to the existing TCP host\:port option. The configuration, startup logic, and documentation have been updated to support this new mode while keeping validation strict to ensure only one listening method is used.

---

### Server listening mode enhancements

* Added support for listening on a Unix socket via the new `socket-path` configuration property in `config.go`, with validation to ensure that only one of `socket-path` or `host:port` is specified. [\[1\]](diffhunk://#diff-d18ceb0046af69e223e4a56710b3682b62d24df872ce326c1ff9db5ff04e10beR34) [\[2\]](diffhunk://#diff-d18ceb0046af69e223e4a56710b3682b62d24df872ce326c1ff9db5ff04e10beR487-R498)
* Modified server startup logic in `glance.go` to handle Unix socket creation and cleanup, and to select between TCP and Unix socket based on configuration. [\[1\]](diffhunk://#diff-cc6a3744cb184c0fafd7bc4e41483535660c01dd92f328472c2d3b647dfdc3f8L492-R534) [\[2\]](diffhunk://#diff-cc6a3744cb184c0fafd7bc4e41483535660c01dd92f328472c2d3b647dfdc3f8R9-R11)

---

### Documentation updates

* Updated `docs/configuration.md` to document the new `socket-path` option.
* Added configuration examples for both TCP and Unix socket modes, and clarified that only one mode may be enabled at a time. [\[1\]](diffhunk://#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180L279-R302) [\[2\]](diffhunk://#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180R313-R323)

---

### Why Unix sockets?

The main motivation for this feature is to simplify reverse proxy setups. For example, a proxy (such as Caddy or Nginx) can be configured to talk directly to the server over a Unix domain socket instead of a TCP port. This avoids the need to open or manage additional firewall rules, which is especially convenient in NixOS environments or when running in containers.

More generally, Unix sockets provide:

* No TCP port required → reduced attack surface and simpler firewalling.
* Lower overhead for local communication (no IP stack traversal).
* File system–based access control via standard Unix permissions.
* A good fit for sandboxed/containerized deployments where networking may be restricted.


AWildLeon:
Any Feedback welcome :smile:

PR #825, opened 2025-09-22
[dns-stats] added support for adguard installations inside GL.iNet routers

himazawa:
This PR add support for retrieving DNS stats if `adguard home` is installed inside a `GL.iNet` router since they don't use the common authentication system but instead set a route and put a cookie in front. 
GL.iNet used to have public APIs but recently they removed the documentation so some of I had to retrieve data by inspecting the traffic.

A new optional parameter is introduced in the configuration file: `glient-endpoint` if specified the widget will first login against that endpoint and then send the request to retrieve the `adguard` stats as always.

In this PR I also had to write a native md5crypt implementation since the core doesn't have it, if you want to use an external library instead let me know.

I tested this PR in my local environment and it works properly (tested against a Flint 3)


himazawa:
Any plan to merge this? @svilenmarkov

Issue #535, opened 2025-03-30
Use sprig for custom api templates


st0le:
### Description

I noticed on the custom API handler, all functions are implement within the file.

I'd recommend using this [sprig](https://github.com/Masterminds/sprig) instead so there's more functions available.

For what its worth, sprig is used in oh-my-posh as well and is a very mature library.

not-first:
I second this. I also stumbled upon the allegedly better maintained [sprout](https://github.com/go-sprout/sprout).

svilenmarkov:
Someone from the Discord server already mentioned sprig a while ago and I've been looking at their docs for inspiration, but I haven't straight up added the library for a couple of reasons. 

* I want to keep the templates simple and with idiomatic ways of doing things. The lisp/clojure-style syntax that Go's templates use gets ugly very quickly when surrounded by HTML, so the less code it takes to achieve something, the better. IMO Sprout provides too many "low-level" functions that allow you to fiddle around with minute and unnecessary details, making the templates more imperative, when I want them to be more declarative. I prefer adding new functions as necessary and having them work well with one another. 
* There are name collisions with existing functions, meaning this would be a backwards-incompatible change, which would unnecessarily break a bunch of the existing custom widgets. 
* It's one more place where you have to look at docs. If you're someone who isn't familiar with Go's template syntax and want to make a custom widget, you now have to read Go's docs, gjson's docs, Glance's docs, and Sprout's docs. That's _a lot_ and very cumbersome. 
* It's another dependency, and I'm trying to keep them to a minimum.

ilysuha:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

PR #838, opened 2025-10-04
Add server-side storage support for todo widget

wei840222:
## Overview

This PR enhances the todo widget functionality by adding support for server-side storage, allowing todo tasks to be persisted and synced across different devices and browsers.

### Core Functionality
- Added `storage-type` configuration option for todo widget with two modes:
  - `browser` (default): Tasks stored in browser's localStorage
  - `server`: Tasks stored on server in JSON files
- Implemented server-side storage using `afero` filesystem abstraction
- Added new API endpoints for todo data management:
  - `GET /api/widgets/todo/{id}` - Retrieve todo data
  - `PUT /api/widgets/todo/{id}` - Save todo data

### Configuration
- Added `data-path` server configuration option (default: `./data`)
- Automatic creation of `todos` data directory on startup
- Updated server startup log to display configured data path

### Code Quality
- Converted `Todo()` function to async for better data loading
- Improved code structure with `loadFromStorage()` and `saveToStorage()` helper functions

### Documentation
- Updated `docs/configuration.md` with:
  - New `data-path` property in Server section
  - New `storage-type` property in Todo widget section
  - Docker volume mounting examples for data persistence
  - Usage examples for both storage types

## Files Changed
- `.gitignore` - Added data directory
- `go.mod` & `go.sum` - Added afero dependency
- `internal/glance/config.go` - Added DataPath field
- `internal/glance/glance.go` - Implemented data path initialization and API routes
- `internal/glance/static/js/todo.js` - Refactored with storage abstraction
- `internal/glance/templates/todo.html` - Added storage-type data attribute
- `internal/glance/widget-todo.go` - Implemented server-side storage logic
- `docs/configuration.md` - Updated documentation

## Testing
- [x] Server-side storage creates data directory automatically
- [x] Browser storage maintains backward compatibility
- [x] API endpoints handle GET/PUT requests correctly
- [x] Multiple todo widgets can use different storage types
- [x] Tasks persist across server restarts (server mode)

## Breaking Changes
None. This is a backward-compatible enhancement. Existing todo widgets will continue to use browser storage by default.

## Example Configuration

```yaml
server:
  data-path: ./data  # Optional, defaults to ./data

pages:
  - name: Home
    columns:
      - size: full
        widgets:
          # Browser storage (default, backward compatible)
          - type: todo
            title: My Local Tasks
            
          # Server storage (new feature)
          - type: todo
            title: My Synced Tasks
            storage-type: server
            id: shared-tasks

### CODE REVIEW

**Comment by russssl** [comment]
amazing work

File: internal/glance/glance.go [hunk 1]
```diff
@@ -439,6 +442,23 @@ func (a *application) server() (func() error, func() error) {
...
+	absDataPath, _ = filepath.Abs(a.Config.Server.DataPath)
+
+	fs := afero.NewOsFs()
+	todosDataPath := path.Join(absDataPath, "todos")
+	exist, _ := afero.DirExists(fs, todosDataPath)
+	if !exist {
+		fs.MkdirAll(todosDataPath, os.ModePerm)
+	}
+
+	todoWidget := &todoWidget{storage: afero.NewBasePathFs(fs, todosDataPath)}
```
tinyoverflow: You're creating a second reference to the todo widget, in addition to the widgets.go file. I'd suggest moving that logic into the widgets.go file.

internal/glance/glance.go
```diff
@@ -439,6 +442,23 @@ func (a *application) server() (func() error, func() error) {
...
+	mux.HandleFunc("GET /api/widgets/todo/{id}", todoWidget.handleRequest)
+	mux.HandleFunc("PUT /api/widgets/todo/{id}", todoWidget.handleRequest)
```
tinyoverflow: Doing this here creates an explicit binding to the todo component. It doesn't allow the application to be independent of a single component (todo in this case). I'd like to suggest moving this logic into the widget itself, maybe defined through an interface to standardize this procedure for future usage in other/new components.

internal/glance/glance.go
tinyoverflow: You're creating a second reference to the todo widget, in addition to the widgets.go file. I'd suggest moving that logic into the widgets.go file.

Defining the widgets data folder should imho be dependent on the widgets name and not be freely picked here.

internal/glance/static/js/todo.js
```diff
@@ -37,11 +39,23 @@ function inputMarginAnim(entrance = true) {
...
-function saveToLocalStorage(id, data) {
+async function saveToStorage(id, type, data) {
+    if (type === "server") {
+        await fetch(`/api/widgets/todo/${id}`, {
+            method: "PUT",
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify(data),
+        });
+        return;
+    }
```
tinyoverflow: Maybe this logic could be extracted/abstracted to a common library usable by other widgets to enforce a common standard.

go.mod
```diff
@@ -6,10 +6,11 @@ require (
 	github.com/fsnotify/fsnotify v1.9.0
 	github.com/mmcdole/gofeed v1.3.0
 	github.com/shirou/gopsutil/v4 v4.25.5
+	github.com/spf13/afero v1.15.0
```
tinyoverflow: What specific requirement does the afero package solve in this situation, that the built in go FS library can't? Your chances of getting this merged might increase if you're using the built-in functions, as [the maintainer tries to keep the use of third-party packages to a minimum](https://github.com/glanceapp/glance/issues/535#issuecomment-2796975486).

wei840222:
Response to @tinyoverflow's Review
Thank you for the detailed feedback! I have addressed all of your concerns with a comprehensive refactor. Here's my response to each point:

1. Removed Third-Party Dependencies
✅ Resolved - Completely removed the afero dependency and switched to Go's built-in `os` `path/filepath` packages.

> **Rationale**
As you mentioned, the project should minimize third-party dependencies. The afero package was indeed overkill for this use case, and the standard library provides everything we need.

2. Eliminated Hard-Coded Todo Widget Binding
✅ Resolved - Removed the explicit todo widget instantiation and route registration from `glance.go`.

> **Solution**
Extended the existing `widget` interface with a getHandlerFunc() method
In the server method of `glance.go`, dynamically iterate through `widgetByID` to register API routes for all widgets
Any widget can now register its own API routes by implementing the getHandlerFunc() method.

3. Created Universal Storage Abstraction
✅ Resolved - Built a generic storage system that's not limited to the todo widget.

> **Implementation**
Created file-storage.go with common file storage functions
Data is automatically organized by widget type: {data-path}/{widget-type}/{id}
Frontend WidgetStorage class provides unified browser/server storage interface

4. Improved Architectural Design
✅ Resolved - The application is now completely decoupled from specific widget implementations.

> **Architectural Improvements**
Widgets are responsible for registering their own API routes (via getHandlerFunc())
Generic file storage system available to all widgets
Support for nested widgets (group and split-column widgets) in API route registration.

5. Maintained Backward Compatibility
✅ Ensured - All existing functionality remains unchanged, this is a pure enhancement.

> **Key Technical Improvements**
> * **Dependency Removal** afero → Go standard library
> * **Decoupled Design** Hard-coded binding → Dynamic interface discovery
> * **Generalization** Todo-specific → Available to all widgets
> * **Standardization** Special logic → Standardized interface
>
> This refactor not only addresses all your concerns but also establishes a solid foundation for server-side storage functionality in future widgets. Any new widget can easily implement the getHandlerFunc() method to add its own API endpoints.

Thank you again for the valuable feedback! These suggestions helped me build a much cleaner and more extensible architecture.

internal/glance/static/js/widget-storage.js
```diff
@@ -0,0 +1,81 @@
...
+     * Save data to storage
+     * @param {string} key - The storage key
+     * @param {any} data - The data to save
+     * @returns {Promise<void>}
+     */
+    async save(key, data) {
+        if (this.storageType === "server") {
+            return this.saveToServer(key, data);
+        }
+        return this.saveToBrowser(key, data);
```
tinyoverflow, line 34: You return the result of the underlying method calls, but those methods do not return their corresponding result. In my opinion it would make sense to return the promise here, so the caller can handle any errors on it's own. The localStorage save action can be encapsulated in a Promise to streamline this.

internal/glance/static/js/todo.js
```diff
@@ -116,7 +113,7 @@ function Todo(id) {
     const saveItems = () => {
         if (isDragging) return;
 
-        saveToLocalStorage(
+        storage.save(
             id, items.children.map(item => item.component.serialize())
         );
```
tinyoverflow: We should somehow handle any error in this, after the return issue in the widget-storage.js file has been resolved. Even if it's just an alert. It's not fancy, but to my knowledge this frontend does not provide any notification functionality, like toasts, at the moment.

(Implementing such a notification functionality would be out of the scope of this PR and should be implemented in another PR, imho)

internal/glance/file-storage.go
```diff
@@ -0,0 +1,32 @@
...
+	dataPath = path
+}
+
+func ensureDir(widgetType string) error {
+	return os.MkdirAll(filepath.Join(dataPath, widgetType), 0755)
+}
+
+func loadFile(widgetType, key string) ([]byte, error) {
+	if err := ensureDir(widgetType); err != nil {
+		return nil, err
```
tinyoverflow, line 24:
```suggestion
		return nil, fmt.Errorf("ensure widget directory exists: %w", err)
```

internal/glance/file-storage.go
```diff
@@ -0,0 +1,32 @@
+package glance
+
+import (
+	"os"
```
tinyoverflow, line 5:
```suggestion
	"fmt"
	"os"
```

internal/glance/file-storage.go
```diff
@@ -0,0 +1,32 @@
...
+	if err := ensureDir(widgetType); err != nil {
+		return nil, err
+	}
+
+	return os.ReadFile(filepath.Join(dataPath, widgetType, key))
+}
+
+func saveFile(widgetType, key string, data []byte) error {
+	if err := ensureDir(widgetType); err != nil {
+		return err
```
tinyoverflow, line 32:
```suggestion
		return fmt.Errorf("ensure widget directory exists: %w", err)
```

internal/glance/widget-todo.go
```diff
@@ -1,24 +1,93 @@
...
+	StorageType string        `yaml:"storage-type"`
 }
 
 func (widget *todoWidget) initialize() error {
 	widget.withTitle("To-do").withError(nil)
 
+	switch widget.StorageType {
+	case "server":
+		if widget.TodoID == "" {
+			widget.TodoID = strconv.FormatUint(widget.GetID(), 10)
```
tinyoverflow: This, unfortunately, might lead to negative side effects. The widget ID is based on the order inside the config file and thus may not be the same anymore if the user adds another widget. This would result in a loss of data. We could either fail here or demand a user ID. Failing here if no ID has been set would be no issue, as this is a new feature and can't break existing setups.

Otherwise we could fallback to something like "default", as long as it's properly documented. What do you think?

tinyoverflow (commented):
Thank you very much for all the effort you put into this PR! I really like your ideas!
After reviewing this more carefully, I've found a few more spots about which I'd like to give you my 2 cents and hear your opinion about.

internal/glance/widget-todo.go
wei840222: image

```yaml
pages:
  - name: Home
    # Optionally, if you only have a single page you can hide the desktop navigation for a cleaner look
    # hide-desktop-navigation: true
    columns:
      - size: small
        widgets:
          - type: to-do
            hide-header: true
            # storage-type: server
          - type: to-do
            hide-header: true
            # storage-type: server
```

The current behavior is that if a user has multiple todos but no ID is set, all todos will share the same data in the same localStorage.

We could fallback to "default" to make this behavior consistent, and then mention it in the documentation.

After all, we can't use blanks as filenames for file-storage.

wei840222:
Hi @tinyoverflow 

Thank you for the detailed feedback! I've addressed all your concerns:

✅ Fixed Promise Return Issues
`saveToBrowser()` & `loadFromBrowser()` now properly return Promises.
Added try-catch error handling in `todo.js`

✅ Added Error Handling
In `todo.js` users can see informative alerts when operations fail.
Errors  will also logged to console for debugging.

✅ Enhanced Error Messages
`file-storage.go` Added `fmt.Errorf()` for directory creation errors now include descriptive messages.

✅ Fixed Widget ID Issue
`widget-todo.go` Changed fallback from widget.GetID() to stable "default" string to prevents data loss when configuration changes.

✅ Updated Documentation
Corrected storage api path from `todos/` to `to-do/`
Documented new default behavior and added Best Practices section with ID recommendations.

tinyoverflow (approved):
LGTM, thank you so much!
@svilenmarkov what do you think about this one?

tinyoverflow:
Hey @svilenmarkov,
I'd really like to assist you with this. Is there anything we can do to assist you to get this merged?

satr14washere:
any progress to this?

ph-256:
what happened with this 🤔 is it getting merged

wei840222:
@tinyoverflow @svilenmarkov  resolved conflicts, thx

tinyoverflow:
> @tinyoverflow @svilenmarkov resolved conflicts, thx

Hey @wei840222, I’m sorry for the late reply. Just saw this. 

Unfortunately I’m not affiliated with this project, but I’m a fan. Therefore I did reply to some issues and reviewed PRs. In the past I’ve also got in contact with @svilenmarkov to talk about possibilities to keep this project alive in the short and long term, but we didn’t find an agreement.  

I really appreciate your effort, though. I know just too well how demotivating it can be, when nothing happens with a PR.

russssl:
> > @tinyoverflow @svilenmarkov resolved conflicts, thx
> 
> Hey @wei840222, I’m sorry for the late reply. Just saw this.
> 
> Unfortunately I’m not affiliated with this project, but I’m a fan. Therefore I did reply to some issues and reviewed PRs. In the past I’ve also got in contact with @svilenmarkov to talk about possibilities to keep this project alive in the short and long term, but we didn’t find an agreement.
> 
> I really appreciate your effort, though. I know just too well how demotivating it can be, when nothing happens with a PR.

there is a project called dynacat - actively developed fork of glance. might look into that

PR #847, opened 2025-10-15
fix: uses the correct path for assets (ex: manifest.json )

koalalorenzo:
I have noticed this error when I was inspecting my dashboard. Adding a slash to the base url does not fix it... but this sm0ll sm0ll sm0ll code change does!

@svilenmarkov thank you for making this small but powerful jewel of tool :-) It is super useful!


svilenmarkov (approved):
Thanks for contributing!

Issue #370, opened 2025-02-19
Implement a fallback when SVG not available in selfh.st


leleogere:
### Description

While being available on selfh.st/icons, Immich-Power-Tools does not any icon when specified `sh:immich-power-tools` in the config:
![Image](https://github.com/user-attachments/assets/662d1928-6b24-4f2e-b260-50996614c67d)

The thing is selfh.st/isons does not offer a SVG version for all the referenced apps (eg. immich-power-tools). In that case, it would be interesting to fall back to the `webp` version, or even to the PNG one if it's the only available.

However, I guess that glance does not perform any check about the availability of the icons before displaying them in the dashboard, so feel free to close this if you believe this is out of scope of glance.

svilenmarkov:
By default Glance will try to use the SVG version of the icon, however you can manually tell it to use PNG instead for instances like this where an SVG variant isn't available by appending a `.png` at the end:

```yaml
sh:immich-power-tools.png
```

Admittedly this feature isn't properly documented, I'll get it added.

PR #852, opened 2025-10-22
fix: add .webp as overwrite option for icons + docs

gan-of-culture:
#370 

Allow `.webp` to be used as fallback image format if `.svg` is not available. The real world example is the Tdarr icon for example.


svilenmarkov (approved):
Thanks for contributing!

Issue #856, opened 2025-10-31
ability to customize the dashboard per user


junglebells:
### Description

Not all links are meant for all users. It would be nice to customize the page so that certain users are able to access different links than others.

best case: create user groups with individual users and assign page configuration to each group.

Example: admin has access to all links/ pages. super user/ manager has access to a sub-set, and a regular user has even more restricted access.

A simple config where we can define pages (e.g home.yml, glance.yml) for each authenticated user - would be an excellent start:

`auth:
  secret-key: # this must be set to a random value generated using the secret:make CLI command
  users:
    admin:
      password: 123456
      pages: <list of pages>
    svilen:
      password: 123456
      pages: <list of pages>`

junglebells:
The other option is to specify list of common pages, and then "include" or "exclude" per user.

beanaroo:
This may be related to https://github.com/glanceapp/glance/issues/694

junglebells:
> This may be related to [#694](https://github.com/glanceapp/glance/issues/694)

Correct, we are looking for the same outcome. Would love to have this feature implemented.

abhi-singhs:
I have created a PR https://github.com/glanceapp/glance/pull/858 that solves this.
Please review and let me know.

Issue #694, opened 2025-05-24
Page Specific Authentification


louis-thevenet:
### Description

> What problem would this feature solve?

I'd love to share my Glance instance with people but it has a private page.

It would be great if we could configure a list of allowed users for each page.

> If applicable, what would the configuration for this feature look like?

```yaml
pages:
  - name: Private
    allowed-users:
      - admin
      - friend
    columns: ...
```

krillin666:
Would also love to see this implemented!

beanaroo:
This would be great for hosting both `personal` and `work` dashboards on the same instance of glance.

abhi-singhs:
I have created a PR https://github.com/glanceapp/glance/pull/858 that solves this.
Please review and let me know.

PR #858, opened 2025-11-02
feat: Per-page user access control feature

abhi-singhs:
Solves https://github.com/glanceapp/glance/issues/694 and https://github.com/glanceapp/glance/issues/856.
User-based page access control to the application, allowing pages to specify which users are permitted to view them. It also updates navigation and UI elements to reflect the authenticated user's identity and restricts navigation links to only those pages the user can access.
- Add functions to retrieve username from request and check page access.
- Update page struct to include allowed users.
- Modify template data to include accessible pages for the user.
- Enhance HTML templates to display logged-in username and restrict access to pages based on user permissions.
- Introduce new CSS styles for username display in mobile and site views.

Usage:
```
pages:
  - name: Private
    allowed-users:
      - home
      - work
    columns: ...
```
Note: This is an optional field. If this property is not added then the page is available to all logged-in users. 

Example: 
image

image

`glance.yml`:
```
server:
  assets-path: /app/assets

auth:
  secret-key: ....
  users:
    home:
      password-hash: ....
    work:
      password-hash: ....

theme:
  custom-css-file: /assets/user.css

pages:
  - $include: home.yml
  - $include: admin.yml
  - $include: work.yml
  - $include: misc.yml
```

`home.yml`:
```
- name: Home
  columns:
    - ....
```

`admin.yml`:
```
- name: Admin
  allowed-users:
    - home
  columns:
    - ....
```

`work.yml`:
```
- name: Work
  allowed-users:
    - work
  columns:
    - ....
```

`misc.yml`:
```
- name: Misc
  allowed-users:
    - home
  columns:
    - ....
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -283,6 +283,56 @@ server:
 
 When set to `true`, Glance will use the `X-Forwarded-For` header to determine the original IP address of the request, so make sure that your reverse proxy is correctly configured to send that header.
 
+### Restricting pages to specific users
+
+You can restrict access to specific pages by using the `allowed-users` property on a page. When this property is set, only the listed users will be able to access that page. Users not in the list will receive a 403 Forbidden error when attempting to access the page. Example:
+
+```yaml
+auth:
+  secret-key: # generated secret key
+  users:
+    admin:
+      password: admin123
+    john:
+      password: john123
+    jane:
+      password: jane123
+
+pages:
+  - name: Home
+    # No allowed-users property means all authenticated users can access this page
+    columns:
+      - size: full
+        widgets:
+          - type: rss
+
+  - name: Admin
+    allowed-users:
+      - admin
+    columns:
+      - size: full
+        widgets:
+          - type: monitor
+
+  - name: Work
+    allowed-users:
+      - john
+      - jane
+    columns:
+      - size: full
+        widgets:
+          - type: calendar
+```
+
+In this example:
+- The "Home" page is accessible by all authenticated users (admin, john, and jane)
+- The "Admin" page is only accessible by the `admin` user
+- The "Work" page is accessible by both `john` and `jane`
+
+> [!NOTE]
+>
+> If a page does not have the `allowed-users` property set, all authenticated users will be able to access it. If you want to restrict access to all pages, you must explicitly set the `allowed-users` property on each page.
+
 ## Server
 Server configuration is done through a top level `server` property. Example:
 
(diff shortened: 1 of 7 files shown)


junglebells:
@abhi-singhs - this looks great to me, would love to see it implemented. Thanks for taking it on.

MrNavaStar:
Would love to see this implemented, I have a few users who want a custom dashboard and this would allow me to do that for them!

junglebells:
@svilenmarkov could you please consider adding this feature in the application?

please let us know if you need more people to vouch for it or more use cases. 

Thanks so much

lalishansh:
Hi @abhi-singhs, are you using these commits ??
If yes how are you, especially public and admin ones !
Thanks.
(Note: i use Nix-OS so can override sources)
Edit: I'm just getting started with glance !

abhi-singhs:
Yep @lalishansh 
I am using this as my home server dashboard on a Pi cluster.

cedhuf:
Hey folks,
Commig from #905. I’m using TinyAuth for now, which works nicely. Really love this PR — per-page user access control would be super useful!

@bluewalk also noticed [Dynacat](https://github.com/Panonim/dynacat) a fork of Glance that seems a bit more active. Maybe it’s worth checking there too for features like this or some of the pending PRs.

Thanks @abhi-singhs for tackling this, really appreciated!

[Edit] The [Dynacat](https://github.com/Panonim/dynacat) developer already mentioned on the other issue that he’s open to exploring these additions! Maybe this PR could be considered there as well.

RexWzh:
Thanks for working on per-page access control. I have a related use case that the current `allowed-users` proposal does not appear to cover: **optional login on one Glance instance**.

A visitor should be able to open explicitly public pages without a login redirect, see a Login action, and sign in to reveal private pages in the same site's navigation. For example, a public home/status page and a private account/server page should coexist under one origin. This is also the guest/logged-in distinction raised earlier in this thread; I'm adding concrete acceptance criteria rather than opening a duplicate issue.

Suggested contract:
- Page visibility is explicit (`public` / authenticated, defaulting to today's protected behavior when top-level `auth` exists). Existing configs must not become public after upgrade.
- Guest navigation contains only public pages; authenticated navigation includes pages allowed for that user. Logout returns to the guest view.
- Authorization is enforced server-side for direct page URLs **and** page-content/API endpoints, not just by hiding links or widgets in HTML/JS. Private content/metadata cannot leak through JSON, search, assets, or shared caches; private responses should not be publicly cached.
- The same existing login/session mechanism is reused; no second instance, hostname, or independent account store is required.

Would you consider supporting guests in this PR, or tracking that part as a follow-up? I can help test the guest / authenticated / logout / direct-URL matrix against a candidate implementation.

PR #873, opened 2025-11-19
Ability to login on Enter keypress on inputs

TabulateJarl8:
Hey! First of all, thanks for all the work being done on this!

I was setting up Glance and noticed something slightly annoying for me, so I figured that I would add it since it's only a few lines of code. Currently, you can't press enter to submit the login form, so I added two keydown event listeners on the username and password inputs. 

Let me know what you think of this or if I should do anything differently.


raspberryhead:
nice, love it when i go to make a PR and someone's done it already.

PR #916, opened 2025-12-31
Added sticky desktop navigation

ralphocdol:
```yaml
pages:
  - name: Playground
    sticky-desktop-navigation: true
```
Screencast

### DIFF
--- a/internal/glance/config.go
+++ b/internal/glance/config.go
@@ -75,15 +75,16 @@ type user struct {
 }
 
 type page struct {
-	Title                  string  `yaml:"name"`
-	Slug                   string  `yaml:"slug"`
-	Width                  string  `yaml:"width"`
-	DesktopNavigationWidth string  `yaml:"desktop-navigation-width"`
-	ShowMobileHeader       bool    `yaml:"show-mobile-header"`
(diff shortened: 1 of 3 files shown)


benny01000010:
the code works but
as a feature I just don't like this. screen real estate is valuable and most people's dashboards aren't that long that scrolling to the top is a problem.

also - does it save where you are at if you swap over to another dashboard? or does it start you back up from the top? if that's the case this seems totally useless.

ralphocdol:
> screen real estate is valuable and most people's dashboards aren't that long that scrolling to the top is a problem.

This is opt-in/optional, you don't need to enable it if you don't want/need to.

> also - does it save where you are at if you swap over to another dashboard? or does it start you back up from the top?

Glance by design, reloads the entire page when switching tabs. I'm not sure why you even thought it's a good idea to preserve the position when you switch tabs.


The goal here is to not have to scroll up when you quickly need to switch tabs when you happen to expand a long widget. But then again, this is not enabled by default so you can just forget it even exist.

benny01000010 (approved):
thanks for the context!
I supposed it would've made sense to look at the code before making a snarky comment lol. 
now that I actually have reviewed it everything looks good. you implemented this nicely.

PR #940, opened 2026-01-27
feat: Adding URL Detection support into the search widget

IonasPolitis:
Changed widget-search.go to include a "DetectURL" boolean for the search widget to detect URLs (default is false), and also added to the "search.html" a script to handle it's configuration depending on how the user has set up his widget, detecting if the query is a url, if "https://" scheme is present and adding it if needed, and how to handle the query going forward.

Idk if this is a good way of doing it... It's my first pull request... I just really liked the project and wanted to try and contribute to a feature that I would like to have.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -149,14 +149,14 @@ pages:
     columns:
       - size: full
         widgets:
-          $include: rss.yml
+          - $include: rss.yml
   - name: News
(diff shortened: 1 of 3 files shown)


ralphocdol:
similar to https://github.com/glanceapp/glance/pull/558

Issue #910, opened 2025-12-21
Youtube videos occasionally not loading from some channels


Musakkaf:
### Description

Terminal:

2025/12/21 11:51:33 Starting server on :8080 (base-url: "", assets-path: "")
2025/12/21 11:52:17 ERROR Failed to fetch youtube feed channel=UCXuqSBlHAE6Xw-yeJA0Tunw error="unexpected status code 404 for https://www.youtube.com/feeds/videos.xml?playlist_id=UULFXuqSBlHAE6Xw-yeJA0Tunw, response: <!DOCTYPE html>\n<html lang=en>\n  <meta charset=utf-8>\n  <meta name=viewport content=\"initial-scale=1, minimum-scale=1, width=device-width\">\n  <title>Error 404 (Not Found)!!1</title>\n  <style>\n    *{margin:0;padding:0}html,code{font:15px/22px arial,sans-ser"
2025/12/21 11:52:17 ERROR Failed to fetch youtube feed channel=UCR-DXc1voovS8nhAvccRZhg error="unexpected status code 404 for https://www.youtube.com/feeds/videos.xml?playlist_id=UULFR-DXc1voovS8nhAvccRZhg, response: <!DOCTYPE html>\n<html lang=en>\n  <meta charset=utf-8>\n  <meta name=viewport content=\"initial-scale=1, minimum-scale=1, width=device-width\">\n  <title>Error 404 (Not Found)!!1</title>\n  <style>\n    *{margin:0;padding:0}html,code{font:15px/22px arial,sans-ser"
2025/12/21 11:52:17 ERROR Failed to fetch youtube feed channel=UCsBjURrPoezykLs9EqgamOA error="unexpected status code 404 for https://www.youtube.com/feeds/videos.xml?playlist_id=UULFsBjURrPoezykLs9EqgamOA, response: <!DOCTYPE html>\n<html lang=en>\n  <meta charset=utf-8>\n  <meta name=viewport content=\"initial-scale=1, minimum-scale=1, width=device-width\">\n  <title>Error 404 (Not Found)!!1</title>\n  <style>\n    *{margin:0;padding:0}html,code{font:15px/22px arial,sans-ser"
[...]

PR #953, opened 2026-02-09
#910: when YouTube feed fails display link to the channel

alexey-shary:
Potential "fix" for #910 to display a link to the channel if YouTube feed fails:

image


svilenmarkov:
I don't think this solution makes a lot of sense. I've pushed a change to `main` that would cache the list of videos on success and reuse that list if subsequent requests fail for that channel.

PR #955, opened 2026-02-09
Live-events feature for custom-api widgets and monitoring services

frozendark01:
What was implemented:
Server-side background polling (glance.go):

Added recursive polling that dives into container widgets (groups, split-columns)
Polls monitor and custom-api widgets every 15 seconds
Detects content changes and emits events
Custom-api widget change detection (widget-custom-api.go):

Added PrevCompiledHTML field to track previous content
Compares rendered HTML to detect changes
Emits custom-api:data_changed event when changes detected
Event hub with caching (events.go):

Server-Sent Events broadcast to connected clients
Event caching: Stores recent events for reconnecting clients
Debouncing per widget to prevent event spam
Robust numeric type handling for widget IDs
Browser-side DOM updates (page.js):

Listens for custom-api:data_changed events
Fetches updated widget HTML via /api/widgets/{id}/content/
Replaces widget DOM with new content
Re-initializes all widget setup functions
Widget registration fix:

Widgets inside containers are now registered in widgetByID map
Allows endpoint to fetch child widgets by ID


svilenmarkov:
Similar to #1041.

PR #960, opened 2026-02-14
fixed URL for Selfh.st icons

iamagithublearner:
URL was changed on Selfh.st , so updating the URL to make icons work again via selfh.st

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -149,14 +149,14 @@ pages:
     columns:
       - size: full
         widgets:
-          $include: rss.yml
+          - $include: rss.yml
   - name: News
     columns:
       - size: full
         widgets:
           - type: group
             widgets:
-              $include: rss.yml
+              - $include: rss.yml
               - type: reddit
                 subreddit: news
 ```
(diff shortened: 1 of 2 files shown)


iLuckyW:
I support this. Currently, some applications (like dawarich) display the old icon. This would fix this issue.

svilenmarkov:
As per [the docs](https://www.jsdelivr.com/documentation#id-github):

> Omit the version completely or use "latest" to load the latest one (only works with valid [semver versions](https://semver.org/)): (not recommended for production usage)
> Falls back to the repository's default branch if there are no tagged releases.

Meaning both of these are functionally equivalent:

```
https://cdn.jsdelivr.net/gh/selfhst/icons@main/svg/glance.svg
https://cdn.jsdelivr.net/gh/selfhst/icons/svg/glance.svg
```

Since [the repo](https://github.com/selfhst/icons) doesn't have any tagged releases.

iLuckyW:
I'd like to revisit this.

The two URLs aren't always equivalent in practice. As shown in [selfhst/icons#970](https://github.com/selfhst/icons/issues/970), jsdelivr's CDN can keep serving a stale icon (e.g. dawarich) when the version is omitted, since the repo has no tagged releases for jsdelivr to resolve "latest" against reliably.

So pinning the version here still seems like the safer approach. Would it be possible to reopen this?

svilenmarkov:
> this repo in particular has some issues given it was versioned at one point but no longer is. 

I suppose this may be why there's a discrepancy between the documented behavior and what's actually happening. I've re-added and pushed the change, thanks for looking into it.

Issue #963, opened 2026-02-19
Open video when clicking on thumbnail, not only on video title


antonpetrov145:
### Description

Hello team, I think this could be a nice feature to add

- What problem would this feature solve?
Click on the thumbnail to open the video, not only in the video title

- Are there any potential downsides to this feature?
Not that I am aware of

- If applicable, what would the configuration for this feature look like?
Shouldn't be different than what video widget currently uses

- Are there any existing examples of this feature in other software?
In YouTube itself when user clicks on the thumbnail, the video starts playing

- If applicable, include any external documentation required to implement this feature
No that I can think of

- Anything else you think might be relevant
No

I'd like to implement it myself

Thanks

PR #964, opened 2026-02-20
Click on video thumbnail to open video

antonpetrov145:
Added
```
cursor: pointer
```
to the css for the widget

Added `onclick` event for thumbnail

refers to #963

### DIFF
--- a/internal/glance/static/css/widget-videos.css
+++ b/internal/glance/static/css/widget-videos.css
@@ -3,6 +3,7 @@
     aspect-ratio: 16 / 8.9;
     object-fit: cover;
     border-radius: var(--border-radius) var(--border-radius) 0 0;
+    cursor: pointer;
 }
 
 .video-horizontal-list-thumbnail {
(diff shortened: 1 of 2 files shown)


michaelwschultz:
Yes! This drives me nuts. haha thanks for adding. 🙏  Going to pull this into my fork.

luissantos:
Why didn't you add an a tag around the image or around everything?

antonpetrov145:
@luissantos Mainly because I didn't want to introduce new tags in the html, but wrapping the image with `a` tag was my first intention

ext-luissantos-thermondo:
Why? to avoid breaking something some integration or tests or something else? Anyway great job. Would love to see this merged. I started considering forking this project to get these things merged faster.

antonpetrov145:
@luissantos thank you, because I am not sure if in future there will not be some global style for `a` tags or some other styes or tags that will interfere with this, so IMO better to have the `onclick` action for now.

luissantos:
> @luissantos thank you, because I am not sure if in future there will not be some global style for `a` tags or some other styes or tags that will interfere with this, so IMO better to have the `onclick` action for now.

Thanks. For the explanation.

PR #981, opened 2026-03-20
feat: async per-widget loading with individual placeholders

edgarburgues:
## Summary
- Add opt-in `server.async-loading` config option that changes how the page renders widgets
- When enabled, the page skeleton appears immediately with loading spinners, and each widget fetches its content independently via `/api/widgets/{id}/render`
- Fast widgets appear instantly while slow ones show a spinner without blocking the rest
- Auto-retry (3 attempts with backoff) and fade-in animation on widget load
- When disabled (default), behavior is completely unchanged

## Test plan
- [ ] Set `async-loading: false` (or omit it) and verify the dashboard works exactly as before
- [ ] Set `async-loading: true` and verify the page skeleton renders immediately with spinners
- [ ] Verify widgets load independently and appear with a fade-in animation
- [ ] Simulate a slow/failing widget and verify retry feedback ("Retrying 1/3...")
- [ ] Verify interactive features (search, groups, carousels, popovers) work correctly after async load
- [ ] Verify cached widgets appear near-instantly on page reload

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -301,6 +301,7 @@ server:
 | proxied | boolean | no | false |
 | base-url | string | no | |
 | assets-path | string | no |  |
+| async-loading | boolean | no | false |
 
 #### `host`
 The address which the server will listen on. Setting it to `localhost` means that only the machine that the server is running on will be able to access the dashboard. By default it will listen on all interfaces.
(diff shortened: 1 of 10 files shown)


codedcosmos:
I tried this and it's overall quite great. Though I think it breaks the group widget.
https://github.com/glanceapp/glance/blob/main/docs/configuration.md#group

edgarburgues:
Thanks for the catch @codedcosmos. Confirmed and fixed in 76f49e8.

**Root cause:** when async-loading is enabled, `fetchAndInsertWidget` calls `setupWidgetElement(widgetElement)` with the group widget as the root element. `setupGroups` then did `root.getElementsByClassName("widget-type-group")`, which only returns descendants and never the root itself, so the click handlers for the tabs were never attached and the tabs were inert.

The fix includes the root element in the candidate list when it is itself a `widget-type-group`. Nested groups are already rejected during widget initialization, so there is no risk of double-binding.

Verified end-to-end with a headless browser against a config containing a group with three tabs:
- Before fix: clicking tab[1] left the active tab/panel at index 0 (tabs inert).
- After fix: clicking tab[1] correctly activates tab/panel at index 1.

Non-async path is unchanged (`document.classList` is undefined, so the new branch is skipped).

PR #986, opened 2026-03-29
Updating go and alpine.

WillSkates:
Updating go (1.24.x is EOL) and alpine (only `main` support currently.).

Sources:
https://eol.wiki/go
https://eol.wiki/alpine-linux


svilenmarkov:
Already updated, Mr. Skates.

PR #999, opened 2026-04-11
fix: handle HTTPS Docker endpoints without downgrading to HTTP

### LINKED ISSUE #992: Docker containers widget doesn't allow sock-path with https://
### Description

I'm using [docker-socket-proxy](https://github.com/Tecnativa/docker-socket-proxy) on another host behind Traefik and the widget silently replaces https:// with http:// in the sock-path URL.

Running Glance 0.8.4 in a Docker container. The culprit appears to be [widget-docker-containers.go line 320](https://github.com/glanceapp/glance/blob/6c5b7a3f4cc409e31739b2914bb6636d08299126/internal/glance/widget-docker-containers.go#L320) where http:// is hardcoded.

To reproduce, add the following to glances.yml. The second widget will return `fetching containers: non-200 response status: 404 Not Found`

```yaml
      - type: docker-containers
        hide-by-default: false
        sock-path: http://dockerproxy:2375

      - type: docker-containers
        hide-by-default: false
        sock-path: https://host.example.com:8443
```
The following change seems to resolve the issue for me. Happy to have a go at a PR if useful.

```diff
diff --git a/internal/glance/widget-docker-containers.go b/internal/glance/widget-docker-containers.go
index 702a34b..6a8ed5f 100644
--- a/internal/glance/widget-docker-containers.go
+++ b/internal/glance/widget-docker-containers.go
@@ -286,9 +286,10 @@ func fetchDockerContainersFromSource(
        labelOverrides map[string]map[string]string,
 ) ([]dockerContainerJsonResponse, error) {
        var hostname string
+       var scheme string
 
        var client *http.Client
-       if strings.HasPrefix(source, "tcp://") || strings.HasPrefix(source, "http://") {
+       if strings.HasPrefix(source, "tcp://") || strings.HasPrefix(source, "http://") || strings.HasPrefix(source, "https://") {
                client = &http.Client{}
                parsed, err := url.Parse(source)
                if err != nil {
@@ -299,6 +300,7 @@ func fetchDockerContainersFromSource(
                if port == "" {
                        port = "80"
                }
+               scheme = parsed.Scheme
 
                hostname = parsed.Hostname() + ":" + port
        } else {
@@ -317,7 +319,7 @@ func fetchDockerContainersFromSource(
        ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
        defer cancel()
 
-       request, err := http.NewRequestWithContext(ctx, "GET", "http://"+hostname+"/containers/json?all="+fetchAll, nil)
+       request, err := http.NewRequestWithContext(ctx, "GET", scheme+"://"+hostname+"/containers/json?all="+fetchAll, nil)
        if err != nil {
                return nil, fmt.Errorf("creating request: %w", err)
        }
```

sjhddh:
## Summary
- Fixes Docker containers widget silently downgrading `https://` endpoints to `http://`
- The `https://` prefix was missing from the scheme check, causing HTTPS URLs to fall through to the Unix socket code path
- The request URL was hardcoded with `http://`, so even if parsed correctly, HTTPS would be lost

## Changes
- Added `https://` to the prefix check in `fetchDockerContainersFromSource`
- Captured the actual URL scheme instead of hardcoding `http://`
- Mapped `tcp://` → `http://` (preserving existing behavior)
- Default port is now `443` for HTTPS, `80` for HTTP/TCP

## Test plan
- [x] `go test ./...` passes
- [ ] Manual test with Docker endpoint over HTTPS
- [ ] Verify `tcp://` and `http://` endpoints still work as before

Fixes #992

### DIFF
--- a/internal/glance/widget-docker-containers.go
+++ b/internal/glance/widget-docker-containers.go
@@ -286,22 +286,33 @@ func fetchDockerContainersFromSource(
 	labelOverrides map[string]map[string]string,
 ) ([]dockerContainerJsonResponse, error) {
 	var hostname string
+	var scheme string
 
 	var client *http.Client
-	if strings.HasPrefix(source, "tcp://") || strings.HasPrefix(source, "http://") {
+	if strings.HasPrefix(source, "tcp://") || strings.HasPrefix(source, "http://") || strings.HasPrefix(source, "https://") {
 		client = &http.Client{}
 		parsed, err := url.Parse(source)
 		if err != nil {
 			return nil, fmt.Errorf("parsing URL: %w", err)
 		}
 
+		scheme = parsed.Scheme
+		if scheme == "tcp" {
+			scheme = "http"
+		}
+
 		port := parsed.Port()
 		if port == "" {
-			port = "80"
+			if scheme == "https" {
+				port = "443"
+			} else {
+				port = "80"
+			}
 		}
 
 		hostname = parsed.Hostname() + ":" + port
 	} else {
+		scheme = "http"
 		hostname = "docker"
 		client = &http.Client{
 			Transport: &http.Transport{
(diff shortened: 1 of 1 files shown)


samcro1967:
For anyone following this work: configurable Custom API request timeouts have been implemented and tested in the samcro1967/glance fork in PR #5. The fork adds a `timeout` option to Custom API requests and documents the configuration behavior.

Fork implementation: https://github.com/samcro1967/glance/pull/5

PR #1000, opened 2026-04-16
Add default config and bundle it into Docker image for zero-config de…

RohitChandraM:
…ployment

Adds config/glance.yml with a default dashboard (calendar, clock, markets, search, hacker-news, lobsters, rss, bookmarks, server-stats, releases, to-do) and updates the Dockerfile to copy it into the image so Railway/Render deploys work out of the box without external volume mounts.

https://claude.ai/code/session_01WbRnSWZadXKgYYoGUjQBpY

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,3 +9,4 @@
 !/go.mod
 !/go.sum
 !main.go
+!/config/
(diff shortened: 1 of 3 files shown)


RohitChandraM (commented):
Submit

PR #1004, opened 2026-04-29
Fix weather widget background colour

42willow:
uses theme colours

after:
Screenshot 2026-04-29 at 3 04 53 pm

before:
Screenshot 2026-04-29 at 3 04 20 pm


svilenmarkov:
It's intentionally yellow because it represents the sunrise/sunset, and the sun is yellow. It would be a lot more ambiguous if it followed theme colors.

PR #793, opened 2025-08-13
feat: refresh menu on the navbar and auto-refresh options

andrecalil:
Adds a **refresh** icon to the navbar with options to a single-time refresh and an auto-refresh every 3 minutes.

https://github.com/user-attachments/assets/c7b758e2-d936-48e8-ae09-491d74bba06f

This is my first PR to an open source project. I think I followed all the right steps, but please forgive me if I missed anything important. Feedback and orientation are welcomed.

PR #1005, opened 2026-05-06
feat: per-widget refresh-interval and force-refresh endpoint

### LINKED ISSUE #327: [REQUEST] Autorefresh page/elements content on set interval
Add the ability to autorefresh content on glance page - my use case is I often leave open the glance page and view it from time to time, but I always have to refresh manually. An option to autorefresh the content on the page at a set interval would be nice. Also maybe last refreshed datetime or 'minutes ago' could be added to elements, to keep track of the freshness of information displayed.

jrnxf:
Closes #327 (request).

This is a fresh take on per-widget refresh, written specifically around the two
concerns you raised on that issue and on #793:

> The front-end wasn't built in a way where you can unload/unmount anything, so
> refreshing individual widgets without reloading the entire page isn't possible
> currently.

> The `setupPage` function calls other functions, some of which add event
> listeners to the `window`/`document` and have no way of removing those
> listeners. If you simply call the `setupPage` function over and over, you're
> creating a memory leak…

> I'm not sure how a dedicated refresh button on the page is any different than
> just hitting the refresh button in your browser.

All are addressed below.

## What it does

- New `refresh-interval: <duration>` field on every widget. When set, the
  client polls `GET /api/widgets/{id}/refresh` on that interval and swaps the
  widget's `outerHTML` in place.
- The endpoint at `GET /api/widgets/{widget}/refresh` (the route was already
  reserved with a TODO comment in `glance.go`). `?force=1` bypasses the
  widget's cache; without it, a tick within the cache window just re-renders
  existing data (no upstream call). This is what the existing `cache:` field
  becomes useful for once you have client polling.
- A `validate()` call at config load rejects `refresh-interval` on widget
  types that hold interactive client state, embed external content, or have
  nothing to refetch (`calendar`, `to-do`, `clock`, `iframe`, `html`,
  `search`, `bookmarks`, `group`, `split-column`). Fails fast at config load
  rather than silently ignoring. Minimum 5s as a typo-floor.

### "the frontend wasn't built to unmount" — handled, no global lifecycle refactor

I deliberately did **not** introduce a generic widget lifecycle. Instead:

1. The `setupX()` functions in `page.js`, `popover.js`, `masonry.js` now
   accept an optional `root = document` argument. Existing call sites pass
   nothing and behave identically.
2. A small per-widget cleanup-callback registry (`_cleanupCallbacks` on the
   widget's root element) — only the three setups that bind to
   `window`/`document` or instantiate observers register a callback:
   `setupCarousels` (window resize listener), `setupCollapsibleGrids` and
   `setupMasonries` (`ResizeObserver`). All other setups bind to widget
   descendants and are released naturally when the old DOM is replaced.
3. `runWidgetCleanup(widgetEl)` runs those callbacks before `outerHTML` is
   swapped. The new node has its own fresh registry.
4. `setupDynamicRelativeTime` was the one global timer that captured a stale
   `elements` array — it now queries the document fresh on each tick and is
   initialized only once. Refreshed widgets are picked up automatically.
5. `afterContentReady` fires immediately when called post-init, so refresh-
   path setups don't queue forever.

I verified there's no listener accumulation by leaving a horizontal-cards RSS
widget refreshing for several cycles in a real browser and watching its
`_cleanupCallbacks` count — it stays at 1, not climbing.

### "many times more upstream requests" — the math doesn't actually pencil out that way

The whole mechanism is built around `cache:` being the rate limiter, not the
client interval:

- A client tick inside the cache window calls `requiresUpdate` → false →
  re-renders the existing buffer. **Zero upstream calls.**
- A client tick outside the cache window fires exactly one `update()`,
  serialized by the per-widget mutex (and re-checked inside the lock so two
  concurrent renders can't double-fire).

So with the typical RSS default of `cache: 1h` and `refresh-interval: 1m`,
that's 60 client ticks per hour producing 1 upstream call per hour — the
same call cadence you have today, just delivered to the browser without a
manual reload. The user has to set `cache:` to something tighter to actually
increase upstream traffic, which is the same lever that exists today.

### "only update while in foreground" — done

`visibilitychange` clears all interval handles when the tab is hidden and
reinstalls them on focus, with an immediate refresh if the interval already
elapsed during the hidden period. Tab in the background → zero requests.

### "should be per-widget basis" — done

Opt-in per widget, default off. Disallow list enforced at config load.

### "how is this different from a browser refresh?" — three things

- **Force-bypass the cache.** A browser refresh re-runs the page render but
  `requiresUpdate` short-circuits when the cache is fresh. You can't actually
  pull new data on demand without restarting the server. `?force=1` does.
- **Preserves page state.** Scroll, theme picker, popovers, expanded lists in
  *other* widgets, etc.
- **Per-widget intervals.** The whole point is that monitor / dns-stats /
  server-stats each have their own cadence without a page-wide reload that
  also refetches feeds and other slow widgets.

## Locking

The previously-stubbed `handleWidgetRequest` had a TODO calling out the
locking rework. That's the other half of the backend commit:

- `widgetBase` gets a `sync.Mutex`.
- `renderTemplate` takes the lock (covers all `.Render` paths via the
  shared `templateBuffer`).
- `updateOutdatedWidgets` and `containerWidgetBase._update` lock per-widget
  around `update(ctx)` and re-check `requiresUpdate` inside the lock so two
  concurrent page renders can't both fire the same upstream call.
- The page-level lock around `updateOutdatedWidgets` is removed (and the
  unused `page.mu` field with it).

`Render()` and `update()` don't recursively call each other in any widget
type, so re-entrance isn't a risk. Containers (group, split-column) don't
have their own mutex contention with children — different mutex objects.

## What's not in this PR

- No refresh button UI. Easy follow-up; deliberately out of scope so this PR
  stays focused on the mechanism.
- No persistence of cache across restarts.

## Test notes

- `TestWidgetValidateRefreshInterval` covers the disallow list, the
  minimum-interval floor, and the no-interval baseline.
- `TestRegisterWidgetIncludesContainerChildren` covers recursive
  registration of nested widgets.
- `TestUpdateOutdatedWidgetsDedupesConcurrentCalls` and
  `TestContainerUpdateDedupesConcurrentCalls` fire 10 concurrent renders
  against a counter-stub widget and assert exactly one `update()` fires —
  evidence that the locking scheme actually delivers the API politeness
  claim above.
- Existing tests all pass.
- Manually exercised in browser across: `hacker-news`, `rss` (list +
  horizontal-cards/carousel), `monitor`, `server-stats`, `dns-stats`,
  `custom-api`. Auto-refresh swaps cleanly, popovers and lazy-images keep
  working post-swap, console stays clean across cycles.
- Visibility pause: tab → background → no requests; tab → foreground → fires
  immediate refresh if interval elapsed.

Happy to split the PR, change naming, restructure the cleanup registry, or
rewrite the description — just let me know what fits the project's direction.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -741,7 +741,7 @@ cache: 1d  # 1 day
 > Not all widgets can have their cache duration modified. The calendar and weather widgets update on the hour and this cannot be changed.
 
 #### `refresh-interval`
-How often the widget should automatically refresh in the browser without a full page reload. Same syntax as `cache` (a number followed by `s`, `m`, `h`, or `d`). Each tick force-fetches fresh data, bypassing the widget's cache. The minimum allowed interval is 30 seconds.
+How often the widget should automatically refresh in the browser without a full page reload. Same syntax as `cache` (a number followed by `s`, `m`, `h`, or `d`). Each tick re-renders the widget; if the widget's cache hasn't expired, the existing data is reused, so a tight refresh interval is safe and won't hammer upstream APIs. Use the manual refresh button to force a fresh fetch. The minimum allowed interval is 5 seconds.
 
 ```yaml
 refresh-interval: 1m
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -707,6 +707,7 @@ pages:
 | title-url | string | no |
 | hide-header | boolean | no | false |
 | cache | string | no |
+| refresh-interval | string | no |
 | css-class | string | no |
 
 #### `type`
@@ -738,6 +739,21 @@ cache: 1d  # 1 day
 > [!NOTE]
 >
 > Not all widgets can have their cache duration modified. The calendar and weather widgets update on the hour and this cannot be changed.
+
+#### `refresh-interval`
+How often the widget should automatically refresh in the browser without a full page reload. Same syntax as `cache` (a number followed by `s`, `m`, `h`, or `d`). Each tick force-fetches fresh data, bypassing the widget's cache. The minimum allowed interval is 30 seconds.
+
+```yaml
+refresh-interval: 1m
+refresh-interval: 5m
+refresh-interval: 1h
+```
+
+When the page is hidden (tab in the background), refresh timers pause and resume on visibility — they don't fire while you're not looking.
+
+> [!NOTE]
+>
+> Not all widget types support `refresh-interval`. The following are excluded because they hold interactive client-side state, embed external content, or have no data to refresh: `calendar`, `to-do`, `clock`, `iframe`, `html`, `search`, `bookmarks`, `group`, `split-column`. Setting `refresh-interval` on any of these will fail at config load.
 
 #### `css-class`
 Set custom CSS classes for the specific widget instance.
--- a/internal/glance/config.go
+++ b/internal/glance/config.go
@@ -87,8 +87,7 @@ type page struct {
 		Size    string  `yaml:"size"`
 		Widgets widgets `yaml:"widgets"`
 	} `yaml:"columns"`
-	PrimaryColumnIndex int8       `yaml:"-"`
-	mu                 sync.Mutex `yaml:"-"`
+	PrimaryColumnIndex int8 `yaml:"-"`
 }
 
 func newConfigFromYAML(contents []byte) (*config, error) {
--- a/internal/glance/glance.go
+++ b/internal/glance/glance.go
@@ -230,10 +230,8 @@ func newApplication(c *config) (*application, error) {
 	return app, nil
 }
 
-// registerWidget adds a widget (and any contained children, recursively) to
-// app.widgetByID so the per-widget refresh endpoint can resolve them by ID.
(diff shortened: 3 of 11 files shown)


jrnxf:
@svilenmarkov any chance I could get your eyes on this? Think it would be a big add a lot of people would benefit from! 🍻

Djerbiien:
Any updates on this, should i move to another project if this one is duying ?

jrnxf:
@Djerbiien my read is this is a dead project. no shade to the maintainer, i've let my fair share of projects die lol. I've since moved off of glance to a vibe coded Tanstack Start app that looks really similar to glance but I get to tweak as I like

tn-nickf:
I picked up Glance yesterday because I didn't want to use Node/JS (MirrorMirror, etc) anymore. Then today, I downloaded a Go compiler because I wanted this feature because my vibe coded mess was already done better by someone else :P

svilenmarkov:
Sorry, not currently interested in adding this functionality this on top of the current implementation. Will be added during a proper rewrite.

PR #1022, opened 2026-06-07
feat(reddit): add show-subreddit option to label each post's source

caciolucas:
When the `reddit` widget aggregates multiple subreddits via a multireddit (`subreddit: a+b+c`), all posts are merged into one feed but each post only shows its link domain (e.g. `i.redd.it`), so there is no way to tell which subreddit a post came from.

This adds a `show-subreddit` option that renders `r/<subreddit>` next to each post.

- It is **enabled automatically when `subreddit` is a multireddit** (contains `+`), since the source is otherwise ambiguous.
- It can also be set explicitly with `show-subreddit: true` for single-subreddit widgets.
- Works across all three styles: `vertical-list` (forum-posts), `vertical-cards` and `horizontal-cards`.

### Example
```yaml
- type: reddit
  subreddit: selfhosted+homelab+DataHoarder
  # show-subreddit is implied here because it is a multireddit
```

Each post now shows e.g. `r/homelab`, `r/selfhosted`, `r/DataHoarder` linking to the respective subreddit.

### Changes
- `widget-reddit.go`: parse the per-post `subreddit` field, add the `show-subreddit` option, populate `forumPost.Subreddit` when enabled or when the subreddit is a multireddit.
- `widget-shared.go`: add `Subreddit` to `forumPost`.
- templates: render `r/<subreddit>` in `forum-posts.html`, `reddit-vertical-cards.html`, `reddit-horizontal-cards.html`.
- docs updated.

`go vet` and `gofmt` clean; built and tested locally against a multireddit and a single subreddit with `show-subreddit: true`.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -20,6 +20,9 @@ builds:
       - 386
     goarm:
       - 7
+    ignore:
+      - goos: windows
+        goarch: arm
     ldflags:
       - -s -w -X github.com/glanceapp/glance/internal/glance.buildVersion={{ .Tag }}
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1063,6 +1063,7 @@ Example:
 | style | string | no | vertical-list |
 | show-thumbnails | boolean | no | false |
 | show-flairs | boolean | no | false |
+| show-subreddit | boolean | no | false |
 | limit | integer | no | 15 |
 | collapse-after | integer | no | 5 |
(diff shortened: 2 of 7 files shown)


svilenmarkov:
I do want to make the subreddit available for cases like this, but the issue with adding any more information there is that on mobile with thumbnails enabled, the target URL becomes unclickable:

image

I also don't want to let the list wrap since that ruins the flow and looks disorganized:

image

One alternative is appending the subreddit at the end of the title within `[]`, but that isn't great either. I'm open to other suggestions.

PR #1046, opened 2026-07-29
fix(docker): reflect unhealthy child status

### LINKED ISSUE #1043: Docker Container widget shows overall healthy even if a container in the group is unhealthy
### Description

When grouping multiple containers of a service and one of the child services is in the "unhealthy" state, the overall state of the service is still shown as "OK".

Image

The unhealthy state should be reflected, just like a child container that is not running.

I'm using Glance 0.8.5, installed via docker compose on my home server.

The config for the widget is just this:

```yaml
- name: Home
  columns:
    - size: small
      widgets:
        - type: docker-containers
```

Nameless-Monster-Nerd:
## Summary

- detect Docker's `(unhealthy)` marker in the container status text
- use that health result for both standalone containers and grouped child containers
- add regression coverage for running, unhealthy-running, and unhealthy states

## Testing

- `go test ./internal/glance`
- `git diff --check`

Fixes #1043

### DIFF
--- a/internal/glance/widget-docker-containers.go
+++ b/internal/glance/widget-docker-containers.go
@@ -137,8 +137,12 @@ func (containers dockerContainerList) sortByStateIconThenTitle() {
 	})
 }
 
-func dockerContainerStateToStateIcon(state string) string {
-	switch state {
+func dockerContainerStateToStateIcon(state, status string) string {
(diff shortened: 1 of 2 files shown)


svilenmarkov:
This was already fixed in `dev`, will be added to main.

PR #1051, opened 2026-07-29
Add a same-tab option to Twitch Channels

### LINKED ISSUE #975: Open link in same tab in Twitch Channels widget
### Description

When clicking on each channel they get opened in a seperate tab. With this feature someone could add the same-tab configuration key (similar to the bookmarks widget) to make the links open in the same tab. 

The widget title would open in a seperate tab, because this it is defined in the widgetBase and not in the specific twitch widget. Also fine granularity similar to the bookmarks would not work with the current way channels are being configured.

I would happily implement this feature in a basic way that does not cause backwards incompability and does not mess with the current widgetBase.

Nameless-Monster-Nerd:
Fixes #975.

The Twitch Channels widget now supports same-tab: true. It applies to channel avatar and name links while leaving the widget title and category links opening in a new tab, as requested. The default behavior remains unchanged.

Tests:
- go test ./internal/glance -run TestTwitchChannelsWidgetSameTab -count=1
- go test ./internal/glance -count=1
- go vet ./internal/glance
- git diff --check

AI disclosure: I used Codex to help investigate, implement, and test this change. I reviewed the final diff and test results.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2873,6 +2873,7 @@ Preview:
 | channels | array | yes | |
 | collapse-after | integer | no | 5 |
 | sort-by | string | no | viewers |
+| same-tab | boolean | no | false |
 
 ##### `channels`
 A list of channels to display.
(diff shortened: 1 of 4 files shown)


AWildLeon:
please check before you open a PR. This is a duplicate of #976

Issue #757, opened 2025-07-08
Stock price for London Stocks is in the wrong units


jamespo:
### Description

eg Barclays (BARC.L) is currently displayed as £337.05 which is out by a factor x100. It should be 337.05 pence / p or £3.3705

This is with the latest standalone version self-hosted.

svilenmarkov:
Ah, I see now, thanks for providing that link, I wasn't aware of this. I'll just remove the currency symbol entirely for all LSE stocks.

toastyjc:
Is this planned to be corrected at some stage?
A lot of my stocks are priced in pence as a resident of the UK

Stocks are typically listed as as such:

6,459p (p = pence) or simply 6,459
Or divided by 100 so 6,459 becomes £64.59

A weird oddity with the LSE.

PR #1055, opened 2026-08-03
remove GBP from currencyToSymbol map as units are in pence, not pounds

jamespo:
Fixes glanceapp/glance#757


svilenmarkov:
Thanks for contributing, though I believe the correct fix is to remove `ToUpper` before checking against the `currencyToSymbol` map since the currency gets returned as `GBp`, which specifically indicates that the price is in pence. This fix has already been committed.

jamespo:
Sure, as long as the upper wasn't required for any other currencies that works too.

PR #1056, opened 2026-08-05
docs: add Cloud and SaaS status page

koko3tallah:
## What changed

Adds a copy-ready **Cloud & SaaS status** preconfigured page built from Glance's `custom-api` and `split-column` widgets. The example includes AWS, Cloudflare, GitHub, OpenAI, Anthropic, and Google Cloud, with provider rollups, active incident counts, service-level states, source freshness, and a real local-render screenshot.

The shared YAML anchor keeps the six widgets consistent, and provider slugs can be swapped without changing the template.

## Why

This gives Glance users a practical dependency-status page without API keys or secrets. It uses OutageDeck's read-only public provider endpoints and caches each widget for 10 minutes. The supplied six-widget page therefore makes 36 requests per hour, within the anonymous API's 120-request-per-hour allowance.

Only the clickable provider title links contain campaign attribution parameters; API fetch URLs remain canonical and parameter-free.

## Validation

- `go run . -config /tmp/glance-cloud-status-final.yml config:validate`
- `go test ./...`
- `go vet ./...`
- `go build -o /tmp/glance-outagedeck-check .`
- Loaded the page in a local Glance server and verified that all six cards rendered without widget errors
- Verified all six production endpoints returned HTTP 200 with the expected provider/status/services/incidents schema
- `git diff --check`

## Disclosure

I am contributing this as part of OutageDeck's distribution work, so I have an affiliation with the API used by the example. The page is keyless and read-only, sends no credentials, and uses only fixed HTTPS endpoints shown directly in the config.

This contribution was prepared with AI assistance and then validated locally with the checks above.


ralphocdol:
Wrong repo, this should be PR'd in https://github.com/glanceapp/community-widgets

koko3tallah:
Thanks for the redirect — you're right. I've moved the contribution to glanceapp/community-widgets#303 and adapted it to that repository's widget format. I'm closing this PR so the work is tracked in the right place.

PR #1060, opened 2026-08-16
added support for Tautulli and TrueNAS widgets in the Glance module

samwells2000:
added support for truenas status and tautulli status reporting.


svilenmarkov:
These should be contributed as custom API widgets.

Issue #958, opened 2026-02-12
if A {...} else if A {...} pattern


Goshya:
### Description

I found 'if A {...} else if A {...} pattern inside 'parseCliOptions' function in file [cli.go](https://github.com/glanceapp/glance/blob/6c5b7a3f4cc409e31739b2914bb6636d08299126/internal/glance/cli.go#L92) on line 92.

```
func parseCliOptions() (*cliOptions, error) {
  var args []string

  args = os.Args[1:]
  if len(args) == 0 {
    intent = cliIntentServe
  } else if len(args) == 1 {
    ....
  } else if len(args) == 2 {
    if args[0] == "password:hash" {
      intent = cliIntentPasswordHash
    } else {
      return nil, unknownCommandErr
    }
  } else if len(args) == 2 {            // <=
    if args[0] == "mountpoint:info" {
      intent = cliIntentMountpointInfo
    } else {
      return nil, unknownCommandErr
    }
  } else {
    return nil, unknownCommandErr
  }
}
```
There are 2 identical conditions 'len(args) == 2'. This means that the second 'then' branch will never be executed.

michaelwschultz:
Looks like "mountpoint:info" just needs to be in the list above.

no-hup:
Happy to fix this — plan is to merge the two duplicate `len(args) == 2` branches in `parseCliOptions` into one block so `mountpoint:info` becomes reachable, same as #994 proposed before it was closed. PR incoming against dev.

PR #1065, opened 2026-08-26
fix: make mountpoint:info CLI command reachable

no-hup:
`glance mountpoint:info <path>` is listed in `--help` but always fails with `unknown command`. In `parseCliOptions` there are two identical `else if len(args) == 2` branches — the first (`password:hash`) always wins, so the `mountpoint:info` branch is dead code.

Merged the two-arg commands into one block:

```
$ glance mountpoint:info /
Path: /
FS type: apfs
Used percent: 84.5%
```

`password:hash` behaves the same as before. `go vet` and tests pass.

Fixes #958

### DIFF
--- a/internal/glance/cli.go
+++ b/internal/glance/cli.go
@@ -86,11 +86,7 @@ func parseCliOptions() (*cliOptions, error) {
 	} else if len(args) == 2 {
 		if args[0] == "password:hash" {
 			intent = cliIntentPasswordHash
-		} else {
-			return nil, unknownCommandErr
-		}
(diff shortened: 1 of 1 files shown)


no-hup:
Gentle bump on this one. It's a small +1/-5 fix that makes the `mountpoint:info` command reachable again — a duplicated `else if len(args) == 2` branch in `parseCliOptions` had left it as dead code. Build and tests are green, and it still merges cleanly onto `dev`. Happy to rebase if it's drifted. Thanks!

Issue #1068, opened 2026-09-02
Add a `link` page property for external top-navigation entries (no local content)


framinosona:
## Problem this would solve

Every entry in the top-level `pages:` list currently renders as a local page: it gets a slug, a route, and its own `columns:`/`widgets:` content. There's no way to add a nav item that's simply an external link — one that just navigates the browser to another URL, the way a plain `<a href>` would, without Glance trying to render page content for it.

**Concrete use case**: I run Glance natively on three machines in my homelab (each with its own `docker-containers`/`server-stats` widgets, since those widgets need a local `docker.sock` and local disk — they can't be usefully proxied). I'd like each instance's top nav to link straight to the other instances' real pages, the same way `Home`/`Bookmarks`/etc. already work as nav tabs. Today there's no way to do that — the only workaround is embedding another instance's page in an `iframe` widget on a dedicated page, which works but is a real downgrade: a nested page inside a fixed-height frame, double-applied theme CSS, and (if the parent page is HTTPS while the target isn't) it can trip the browser's mixed-content blocking entirely, requiring an extra reverse-proxy hop just to make the embed same-scheme.

A plain external nav link sidesteps all of that — no iframe, no nested rendering, no mixed-content concern (it's a real top-level navigation, not a same-page embed).

## Proposed solution

Add an optional `link` property to a page entry. When set, that page contributes a normal top-nav tab (same styling, same position in the list) but its `<a>` points straight at the given URL instead of `/{slug}` — no route is registered for it, no content is rendered, clicking it is a real navigation just like any other external link.

```yaml
pages:
  - $include: pages/home.yml
  - $include: pages/bookmarks.yml
  - name: Server A
    slug: server-a
    columns: [...]        # this machine's own real page
  - name: Server B
    link: https://server-b.example.com:8085/
  - name: Server C
    link: https://server-c.example.com:8085/
```

Each of my three Glance instances would carry the *same* page list, with its own machine as a real `$include`d page and the other two as `link` entries — genuinely feels like one app spanning three machines from the nav bar, with zero data-pulling logic anywhere.

## Where this fits in the current code (I read through it before filing this, to save implementation time)

[...]

PR #1070, opened 2026-09-02
Fix pkg/sysinfo mountpoint auto-detection returning nothing in plain containers

framinosona:
Addresses the root cause raised in #1068 — thanks for pointing at `pkg/sysinfo` specifically, that's exactly where it turned out to live.

## The bug

`disk.Partitions(false)` asks gopsutil for "physical" filesystems only — but its notion of physical excludes `overlay`, the storage driver Docker uses for a container's own root filesystem by default. In a plain container with no extra host bind mounts, that leaves auto-detection with nothing to report except a handful of tiny `/etc/hosts`/`/etc/hostname`/`/etc/resolv.conf` config-file bind mounts Docker injects into every container (which callers like `glanceapp/agent`'s default mountpoint list already hide anyway). Net result: an empty `mountpoints` list — not because anything errors, just because nothing survives gopsutil's "physical" filter.

## How I confirmed this, not just guessed it

Companion PR [glanceapp/agent#6](https://github.com/glanceapp/agent/pull/6) fixes a real bug where `LOG_LEVEL=debug` never actually produced output (nothing in that codebase ever configured `log/slog`'s handler to emit Debug-level records) — needed that fixed first to actually *see* what `Collect` was doing instead of guessing from silence.

With that working, I wrote a small standalone probe comparing `disk.Partitions(false)` vs `disk.Partitions(true)` inside a real container:

```
=== disk.Partitions(false) — physical only ===
  device=/dev/mapper/... mountpoint=/etc/resolv.conf fstype=ext4 opts=[rw relatime bind]
  device=/dev/mapper/... mountpoint=/etc/hostname    fstype=ext4 opts=[rw relatime bind]
  device=/dev/mapper/... mountpoint=/etc/hosts       fstype=ext4 opts=[rw relatime bind]

=== disk.Partitions(true) — all ===
  device=overlay mountpoint=/ fstype=overlay opts=[rw relatime]
  device=proc    mountpoint=/proc ...
  (+ the same three /etc/* bind mounts, + sysfs/tmpfs/devpts/cgroup/etc.)
```

The overlay root — the thing you'd actually want disk usage for — simply never appears under `(false)` at all.

## The fix

Switch to `disk.Partitions(true)` (get everything) filtered through a `physicalFilesystemTypes` allowlist that includes `overlay` alongside the usual real disk filesystem types (ext4, xfs, btrfs, ntfs, ...). Chose an allowlist over excluding known virtual types, since virtual filesystem types accumulate over time (new pseudo-fs types get added to kernels) while real disk filesystem types are comparatively stable and easy to enumerate.

## Testing

- `go build ./...` and `go vet ./...` clean across the whole repo (nothing else calls `disk.Partitions` or depends on the old physical-only behavior as far as I could find).
- End-to-end against `glanceapp/agent` (via a local `go.mod replace`, no `agent.yml`, no `MOUNTPOINTS` env — the true zero-config default path): mountpoints now correctly reports `/` with real total/used/percent matching the host's actual disk, where it previously reported nothing.
- No test suite exists for this package currently; happy to add table-driven tests for `isPhysicalFilesystemType` if wanted, wasn't sure of your preferred test file layout for `pkg/`.

### DIFF
--- a/pkg/sysinfo/sysinfo.go
+++ b/pkg/sysinfo/sysinfo.go
@@ -260,9 +260,22 @@ func Collect(req *SystemInfoRequest) (*SystemInfo, []error) {
 	}
 
 	if !req.HideMountpointsByDefault {
-		filesystems, err := disk.Partitions(false)
+		// disk.Partitions(false) asks gopsutil for "physical" filesystems only, but its
+		// notion of physical excludes "overlay" -- the storage driver Docker uses for a
+		// container's own root filesystem by default. In a plain container with no extra
+		// host mounts, that leaves auto-detection with nothing to report except whatever
+		// tiny /etc/* config-file bind mounts Docker injects (which get hidden by callers
+		// like glanceapp/agent's default container mountpoint list anyway), so the net
+		// result is an empty mountpoints list even though there's a real, meaningful disk
+		// behind the overlay root worth reporting on. disk.Partitions(true) plus an
+		// allowlist of genuinely disk-backed filesystem types (keeping "overlay") reports
+		// that root correctly while still excluding proc/sysfs/tmpfs/etc noise.
+		filesystems, err := disk.Partitions(true)
 		if err == nil {
 			for _, fs := range filesystems {
+				if !isPhysicalFilesystemType(fs.Fstype) {
+					continue
+				}
 				addMountpointInfo(fs.Mountpoint, req.Mountpoints[fs.Mountpoint])
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


samcro1967:
A data point for this fix: the mountpoint auto-detection change has been incorporated and tested in the samcro1967/glance fork, including containerized Glance operation where automatic mountpoint discovery is particularly important.

Relevant fork integration:
https://github.com/samcro1967/glance/pull/95

PR #1076, opened 2026-09-07
Make glance.hide case insensitive

NateyB:
(& trim whitespace)

The `glance.hide` label is currently case-sensitive. Podman-compose serializes `glance.hide: true` to `glance.hide = "True"` for some reason (possibly a Python thing). While this should probably be considered bad behavior by podman-compose, case-insensitivity on glance's side also makes sense.

Tested with the docker build.


NateyB:
Ah, hold on. This also affects the header use-case which will always fail. Give me a moment.

NateyB:
Fixed!

